PR #5, opened 2019-11-12
Issue #4 - Clean up error handling, make sure disconnect is called co…

nemccarthy:
…rrectly after write. Improve retry logic to also include retrying if the connection attempt fails. Finally logging has been improved.

PR #6, opened 2019-11-18
Improve security - Add password support for switchbots

nemccarthy:
This PR builds on previous improvements to allow for password support.

The password can be setup in the switchbot application on iPhone or Android by;

1. Click the settings icon next to the switchbot
2. Click Password, Mode settings
3. Click on password and set a password using ASCII characters 

Then pass the password into pySwitchbot as a string.


Danielhiversen:
You have mixed in some changes from  #5

PR #8, opened 2019-12-26
Added password support from nemccarthy

magnussand:
This is mearly a manually patched version from nemccarthy's pull request in order to get password support.

### DIFF
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -33,7 +33,10 @@ class Switchbot:
         self._mac = mac
         self._device = None
         self._retry_count = retry_count
-        self._password_encoded = self._passwordcrc(password)
+        if password is None or password == "":
+            self._password_encoded = None
+        else:
+            self._password_encoded = '%x' % (binascii.crc32(password.encode('ascii')) & 0xffffffff)
 
     def _connect(self) -> None:
         if self._device is not None:
@@ -58,11 +61,6 @@ class Switchbot:
             _LOGGER.warning("Error disconnecting from Switchbot.", exc_info=True)
         finally:
             self._device = None
-
-    def _passwordcrc(self, password) -> str:
-        if password is None or password == "":
-            return None
-        return '%x' % (binascii.crc32(password.encode('ascii')) & 0xffffffff)
 
     def _commandkey(self, key) -> str:
         key = ""
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -63,7 +63,6 @@ class Switchbot:
             self._device = None
 
     def _commandkey(self, key) -> str:
-        key = ""
         key_suffix = PRESS_KEY_SUFFIX
         if key == ON_KEY:
             key_suffix = ON_KEY_SUFFIX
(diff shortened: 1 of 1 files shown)


switchbot/__init__.py
```diff
@@ -51,6 +62,19 @@ def _disconnect(self) -> None:
         finally:
             self._device = None
 
+    def _commandkey(self, key) -> str:
+        key = ""
```
Danielhiversen: ??

switchbot/__init__.py
```diff
@@ -51,6 +62,19 @@ def _disconnect(self) -> None:
...
+        key_suffix = PRESS_KEY_SUFFIX
+        if key == ON_KEY:
+            key_suffix = ON_KEY_SUFFIX
+        elif key == OFF_KEY:
+            key_suffix = OFF_KEY_SUFFIX
+        if self._password_encoded is not None:
+            key = KEY_PASSWORD_PREFIX + self._password_encoded + key_suffix
+        else:
+            key = KEY_PREFIX + key_suffix
+        return key
```
Danielhiversen: Why is the key modified if no password is set?

Danielhiversen:
@nemccarthy

switchbot/__init__.py
magnussand: Yes that is strange.

I'll look into it and get back

switchbot/__init__.py
magnussand: I'll get back to you when I've tested more

magnussand: Good catch.

I've removed it and committed. Have tested, but note that the switchbot need to be set in switch mode or press mode from the app in order to use the press, turn_on and turn_off methods.

switchbot/__init__.py
```diff
@@ -51,6 +62,18 @@ def _disconnect(self) -> None:
         finally:
             self._device = None
 
+    def _commandkey(self, key) -> str:
+        key_suffix = PRESS_KEY_SUFFIX
```
Danielhiversen: I think this is better
```

       if self._password_encoded is None:
             return key

```

magnussand: Pushed new version with proposed change

switchbot/__init__.py
```diff
@@ -12,20 +12,31 @@
 UUID = "cba20d00-224d-11e6-9fb8-0002a5d5c51b"
 HANDLE = "cba20002-224d-11e6-9fb8-0002a5d5c51b"
 
+KEY_PREFIX = "5701"
```
Danielhiversen: Unused

switchbot/__init__.py
```diff
@@ -66,9 +87,11 @@ def _writekey(self, key) -> bool:
 
     def _sendcommand(self, key, retry) -> bool:
         send_success = False
+        command = self._commandkey(key)
+        _LOGGER.info("Sending command to switchbot %s", command)
```
Danielhiversen: Change to debug

Danielhiversen:
Thanks

magnussand:
You're welcome!

I need this feature in Home Assistant since I don't want my neighbors to
easily be able to switch off my heating.

Are you the maintainer of the integration with HA?

Regards
Magnus

PR #15, opened 2020-12-19
feature: Enable choice of bluetooth interface in the host

moretea:
Thanks for this library!

This PR enables downstream users (I'm looking at HASS) to pick which one of their bluetooth interfaces should be used.


Danielhiversen (approved):

Danielhiversen:
Thanks

Issue #16, opened 2020-12-29
Feature request: implement discovery of devices


moretea:
There exists a nodejs library (https://github.com/futomi/node-switchbot) that supports discovering SwitchBot devices.

To be able to automatically discover SwitchBot devices from Home Assistant, we should be able to discover SwitchBot devices too.

PR #23, opened 2021-06-08
Add Class to discover all switchbots, all curtain bots or all bot (woHand)

RenierM26:
-Added discovery method. 
-All updates are now handled by single advertisement scan.

This should allow discovery and coordinator based updates in hass integration.

model number (first byte of 16b service data): 
"c" = Curtain
"H" = WoHand or otherwise known as bot. 
"T" = woSensorTH or otherwise known as temp/humidity sensor. (I don't have one and can't test the output)

Returns python dict with all device data of type with single scan.
Example of single curtain (json formated - easier to format in notepad ++):

	"fd8082999999": {
		"mac_address": "fd:80:82:99:99:99",
		"Flags": "06",
		"Manufacturer": "5900fd8082c51181",
		"Complete 128b Services": "cba20d00-224d-11e6-9fb8-0002a5d5c51b",
		"data": {
			"calibration": true,
			"battery": 57,
			"position": 98,
			"lightLevel": 2,
			"rssi": -68
		},
		"model": "c",
		"modelName": "WoCurtain"
	},

Returns python dict with all device data of type with single scan.
Example of returned data for switchbot bot(json formatted - easier to format in notepad ++):

	"e78943999999": {
		"mac_address": "e7:89:43:99:99:99",
		"Flags": "06",
		"Manufacturer": "5900e78943d9fe7c",
		"Complete 128b Services": "cba20d00-224d-11e6-9fb8-0002a5d5c51b",
		"data": {
			"switchMode": true,
			"isOn": true,
			"battery": 91,
			"rssi": -71
		},
		"model": "H",
		"modelName": "WoHand"
	},

### DIFF
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -269,14 +269,14 @@ class SwitchbotDevice:
         send_success = False
         command = self._commandkey(key)
         _LOGGER.debug("Sending command to switchbot %s", command)
-        try:
-            with CONNECT_LOCK:
+        with CONNECT_LOCK:
+            try:
                 self._connect()
                 send_success = self._writekey(command)
-        except bluepy.btle.BTLEException:
-            _LOGGER.warning("Error talking to Switchbot", exc_info=True)
-        finally:
-            self._disconnect()
+            except bluepy.btle.BTLEException:
+                _LOGGER.warning("Error talking to Switchbot", exc_info=True)
+            finally:
+                self._disconnect()
         if send_success:
             return True
         if retry < 1:
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -272,7 +272,7 @@ class SwitchbotDevice:
 
     def get_device_data(self, mac) -> dict:
         """Return data for specific device."""
-        if not self._switchbot_device_data:
+        if not self._all_services_data:
             self.discover()
 
         for item in self._all_services_data:
@@ -292,9 +292,9 @@ class Switchbot(SwitchbotDevice):
 
     def update(self, scan_timeout=5) -> None:
         """Update mode, battery percent and state of device."""
-        _scan_btle_adv = self.discover(scan_timeout=scan_timeout)
+        _success = self.discover(scan_timeout=scan_timeout)
 
-        if _scan_btle_adv:
+        if _success:
             self.get_device_data(self._mac)
 
     def turn_on(self) -> bool:
@@ -365,9 +365,9 @@ class SwitchbotCurtain(SwitchbotDevice):
 
     def update(self, scan_timeout=5) -> None:
         """Update position, battery percent and light level of device."""
-        _scan_btle_adv = self.discover(scan_timeout=scan_timeout)
+        _success = self.discover(scan_timeout=scan_timeout)
 
-        if _scan_btle_adv:
+        if _success:
             self.get_device_data(self._mac)
 
     def get_position(self) -> int:
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -45,9 +45,6 @@ class SwitchbotDevices:
         returns after the given timeout period in seconds."""
         devices = None
 
-        waiting_time = self._time_between_update_command - time.time()
-        if waiting_time > 0:
-            time.sleep(waiting_time)
         try:
             devices = bluepy.btle.Scanner().scan(scan_timeout)
 
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -107,6 +107,8 @@ class SwitchbotDevices:
 
                 bot_sensors[item]["battery_percent"] = _sensor_data[2] & 0b01111111
 
+        return bot_sensors
+
     def all_curtains(self) -> dict:
         """Return all bots with their data."""
         self.discover()
@@ -118,12 +120,17 @@ class SwitchbotDevices:
                 _sensor_data = self._services_data[item]["16b Service Data"]
                 _sensor_data = binascii.unhexlify(_sensor_data.encode())
 
-                curtain_sensors[item]['calibrated'] = _sensor_data[1] & 0b01000000
-                curtain_sensors[item]['battery_percent'] = _sensor_data[2] & 0b01111111
+                curtain_sensors[item]["calibrated"] = _sensor_data[1] & 0b01000000
+                curtain_sensors[item]["battery_percent"] = _sensor_data[2] & 0b01111111
                 _position = max(min(_sensor_data[3] & 0b01111111, 100), 0)
-                curtain_sensors[item]['pos'] = (100 - _position) if self._reverse else _position
-                curtain_sensors[item]['light_level'] = (_sensor_data[4] >> 4) & 0b00001111  # light sensor level (1-10)
+                curtain_sensors[item]["pos"] = (
+                    (100 - _position) if self._reverse else _position
+                )
+                curtain_sensors[item]["light_level"] = (
+                    _sensor_data[4] >> 4
+                ) & 0b00001111  # light sensor level (1-10)
 
+        return curtain_sensors
 
 
 class SwitchbotDevice:
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -289,13 +289,12 @@ class Switchbot(SwitchbotDevice):
         """Switchbot Bot/WoHand constructor."""
         super().__init__(*args, **kwargs)
         self._inverse = kwargs.pop("inverse_mode", False)
-        self._switchbot_device_data = {}
 
     def update(self, scan_timeout=5) -> None:
         """Update mode, battery percent and state of device."""
-        self._switchbot_device_data = self.discover(scan_timeout=scan_timeout)
+        _success = self.discover(scan_timeout=scan_timeout)
 
-        if self._switchbot_device_data:
+        if _success:
             self.get_device_data(self._mac)
 
     def turn_on(self) -> bool:
@@ -345,7 +344,6 @@ class SwitchbotCurtain(SwitchbotDevice):
 
         super().__init__(*args, **kwargs)
         self._reverse = kwargs.pop("reverse_mode", True)
-        self._switchbot_device_data = {}
 
     def open(self) -> bool:
         """Send open command."""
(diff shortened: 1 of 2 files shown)


RenierM26:
Should close #16

RenierM26:
Hi @Danielhiversen,

I'm done with this pull request. Could you please review if you get a chance? 

Thank you.

switchbot/__init__.py
```diff
@@ -220,42 +313,47 @@ def press(self) -> bool:
...
+        # To get actual position call update() first.
+        if not self._switchbot_device_data:
+            self.update()
+        return self._switchbot_device_data["data"]["switchMode"]
 
     def is_on(self) -> bool:
-        """Return switch state from cache.
-        Run update first."""
-        return self._is_on
+        """Return switch state from cache."""
```
Danielhiversen, line 323: I would not expect `is_on` to connect to the device, but return the latest state

switchbot/__init__.py
```diff
@@ -93,119 +149,156 @@ def _writekey(self, key) -> bool:
...
-                        service_data = binascii.unhexlify(service_data)
+                        _model = binascii.unhexlify(value[4:6]).decode()
+                        if _model == "H":
+                            self._all_services_data[dev_id]["data"] = _process_wohand(
+                                value[4:]
+                            )
+                            self._all_services_data[dev_id]["data"]["rssi"] = dev.rssi
+                            self._all_services_data[dev_id]["model"] = _model
+                            self._all_services_data[dev_id]["modelName"] = "WoHand"
+                        if _model == "c":
```
Danielhiversen, line 224:
```suggestion
                        elif _model == "c":
```

switchbot/__init__.py
```diff
@@ -93,119 +149,156 @@ def _writekey(self, key) -> bool:
...
+                            self._all_services_data[dev_id][
+                                "data"
+                            ] = _process_wocurtain(value[4:])
+                            self._all_services_data[dev_id]["data"]["rssi"] = dev.rssi
+                            self._all_services_data[dev_id]["model"] = _model
+                            self._all_services_data[dev_id]["modelName"] = "WoCurtain"
+                        if _model == "T":
```
Danielhiversen, line 231:
```suggestion
                        elif _model == "T":
```

switchbot/__init__.py
```diff
@@ -93,119 +149,156 @@ def _writekey(self, key) -> bool:
...
-            "Cannot connect to Switchbot. Retrying (remaining: %d)...", retry
-        )
+        _LOGGER.warning("Cannot connect to Switchbot. Retrying (remaining: %d)", retry)
         time.sleep(DEFAULT_RETRY_TIMEOUT)
         return self._sendcommand(key, retry - 1)
 
-    def get_servicedata(self, retry=DEFAULT_RETRY_COUNT, scan_timeout=5) -> bytearray:
-        """Get BTLE 16b Service Data,
-        returns after the given timeout period in seconds."""
+    def discover(self, retry=DEFAULT_RETRY_COUNT, scan_timeout=5) -> dict:
```
Danielhiversen, line 184:
```suggestion
    def discover(self, retry=DEFAULT_RETRY_COUNT, scan_timeout=5) -> dict | None:
```

Danielhiversen (changes requested):
Some initial comments.

@Jan108  Any comments? `time_between_update_command ` will be removed

switchbot/__init__.py
```diff
@@ -28,23 +28,79 @@
...
+    return _wosensorth_data
 
 
 class SwitchbotDevice:
-    # pylint: disable=too-few-public-methods
-    # pylint: disable=too-many-instance-attributes
     """Base Representation of a Switchbot Device."""
 
-    def __init__(self, mac, password=None, interface=None, **kwargs) -> None:
+    def __init__(self, mac=None, password=None, interface=None, **kwargs) -> None:
```
Danielhiversen, line 96: When is `mac` not required?

Danielhiversen:
For later, please do not include any formatting changes. Makes it difficult to review and see actual changes.

switchbot/__init__.py
```diff
@@ -93,119 +149,156 @@ def _writekey(self, key) -> bool:
...
-
-        for device in devices:
-            if self._mac.lower() == device.addr.lower():
-                for (adtype, _, value) in device.getScanData():
+            return self.discover(retry - 1, scan_timeout)
+
+        for dev in devices:
+            if dev.getValueText(7) == UUID:
+                dev_id = dev.addr.replace(":", "")
+                self._all_services_data[dev_id] = {}
```
Danielhiversen, line 212: The SwitchbotDevice class should only keep information about one device

switchbot/__init__.py
RenierM26: https://github.com/Danielhiversen/pySwitchbot/blob/9044c5c86d27e56801ba2a085ac9c10315824eca/switchbot/__init__.py#L190

Scanning for advertisement data returns all devices with one scan. The "discover" method returns data for all devices in one 5 second scan.  (Why do multiple scans when one is all that is needed?)

https://github.com/Danielhiversen/pySwitchbot/blob/17f053fcdfce4d9ff88a173e13c92edef31c633c/switchbot/__init__.py#L184
switchbot.SwitchbotDevice().discover() -> returns data for all devices and stores it in self._all_services_data = {}

https://github.com/Danielhiversen/pySwitchbot/blob/17f053fcdfce4d9ff88a173e13c92edef31c633c/switchbot/__init__.py#L276
The is_on fuction returns self._switchbot_device_data["data"]["isOn"] from cache.

Apologies for the long explanation.

switchbot/__init__.py
RenierM26: if..elif..else doesn't make sense for these as there might be a new model that we don't account for. The service data might then be invalid for the last else.

switchbot/__init__.py
RenierM26: if..elif..else doesn't make sense for these as there might be a new model that we don't account for. The service data might then be invalid for the last else.

switchbot/__init__.py
RenierM26: For the discovery function.

SwitchbotDevice().discover() uses btle advertisements and doesn't require a MAC address. It will return all device advertisements for devices in range. (With a single scan)

The uuid is used to identify switchbot devices.

switchbot/__init__.py
RenierM26: I could split it up in different classes? Considering a singe scan returns all data I thought this is the most efficient way of doing it without breaking any 3rd party integrations that makes use of this api. 

Bluetooth adapters can't do many operations at a time and need to be used sparingly. (especially on low power/low cost devices like raspberry pi)

RenierM26:
I have a working prototype Hass integration that makes use of a coordinator to update states of all devices (I have >8 switchbots) and it updates without any btle errors.

This might help to indicate the logic behind the changes?

https://github.com/RenierM26/ha-switchbot-curtain/tree/main/custom_components/switchbot-curtain

Once I've managed to add "send commands" to the coordinator I will submit pull requests on hass. BLE is a real pain to figure out ways to properly time all adapter commands. (you can literally only issue 1 command at a time with the pygatt based python libraries)

switchbot/__init__.py
Danielhiversen: The last `else` could be `elif _model != "T"` ?

switchbot/__init__.py
Danielhiversen: Great to reduce the number of scans.
I do not think it then make sense to have the discover function as part of the SwitchbotDevice

Danielhiversen: This release will anyway have breaking changes

switchbot/__init__.py
```diff
@@ -220,42 +313,47 @@ def press(self) -> bool:
...
+        return self._switchbot_device_data["data"]["switchMode"]
 
     def is_on(self) -> bool:
-        """Return switch state from cache.
-        Run update first."""
-        return self._is_on
+        """Return switch state from cache."""
+        # To get actual position call update() first.
+        if not self._switchbot_device_data:
+            self.update()
```
Danielhiversen, line 326:
```suggestion
            return None
```
Do not want is_on to trigger any scan

switchbot/__init__.py
Danielhiversen: You never update the mac address later?
So `_connect()` and `_writekey()` will fail?

Danielhiversen:
I think you should make a DataHandler class with functions:
discover -> returns list of Switchbot and SwitchbotCurtain
get_bots -> returns list of Switchbot
get_curtains -> returns list of SwitchbotCurtain


Would that make sense for you?

switchbot/__init__.py
RenierM26: Done

RenierM26:
> I think you should make a DataHandler class with functions:
> discover -> returns list of Switchbot and SwitchbotCurtain
> get_bots -> returns list of Switchbot
> get_curtains -> returns list of SwitchbotCurtain
> 
> Would that make sense for you?

No problem. Created a new class for these and modified method in SwitchbotDevice for single MAC scan.

switchbot/__init__.py
RenierM26: I have moved discovery to a new Class.

switchbot/__init__.py
RenierM26: "mac" is a requirement again.

switchbot/__init__.py
RenierM26: Thanks for the idea. Ended up using:

else: 
  continue

Jan108:
The `time_between_update_command` prevented the reading of false data from the curtain (I can only asume it's a problem for all), this happens if you read the service data while the curtain is moving. The position attribute isn't correct at this moment. 

When with the new code this is no longer a problem we can remove it. But if this still can happen, I would say the library should take care of this and not hass or the libaray user.

RenierM26:
> The `time_between_update_command` prevented the reading of false data from the curtain (I can only asume it's a problem for all), this happens if you read the service data while the curtain is moving. The position attribute isn't correct at this moment.
> 
> When with the new code this is no longer a problem we can remove it. But if this still can happen, I would say the library should take care of this and not hass or the libaray user.

Hi @Jan108 and @Danielhiversen ,

Pretty sure I solved this problem with the binary AND which I found in the node-switchbot code. (Pesky things these binary trees)

Just busy testing a few conditions to confirm.

RenierM26:
Hi @Jan108 and @Danielhiversen,

If reading of service data occur while the curtain is moving, it literately returns the curtain position at that time.  Battery and other data is also returned correctly.

RenierM26:
Hi @Danielhiversen,

Think I addressed all of your suggestions. Please have a look when you get a chance?

Thank you.

Danielhiversen (approved):
Great. Thanks 👍 

I do not think GetSwitchbotDevices is a very good name, but I do not have any better suggestion :)

88fingerslukee:
> Great. Thanks 👍
> 
> I do not think GetSwitchbotDevices is a very good name, but I do not have any better suggestion :)

SwitchbotDeviceDiscovery?

PR #28, opened 2021-10-09
Fix issue with SB password issue and Temp issue.

RenierM26:
-Revert master branch to V0.11.0 release. (Latest master runs scanner in separate thread, this causes memory leak on hassio leading to supervisor restarts every 2nd day.)
-Add WoSensorTH key fix. (From Master)
**to do **:
Fix password issue on discovery.

### DIFF
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 import binascii
 import logging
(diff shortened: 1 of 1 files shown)


RenierM26:
Hi @Danielhiversen,

Could you please have a look at this pull request when you get a chance? It fixes all the issues reported on hassio core with the new integration.

I have tested it locally.

PR #32, opened 2021-10-30
Change library from bluepy to Bleak (with async)

RenierM26:
Bluepy development seems to be stale and plagued by custom code to underlying bluez component.

It might be worth it to rather use something like Bleak? The development is active and it's an asyncio library that seems to work much better.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ setup(
     name = 'PySwitchbot',
     packages = ['switchbot'],
     install_requires=['bleak'],
-    version = '0.12.0',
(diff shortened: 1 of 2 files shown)


Danielhiversen:
Agree that Bleak is much better to work with.
You have my support on this 👍

RenierM26:
Hi @Danielhiversen,

The bleak conversion to pyswitchbot is finally ready for review.

Danielhiversen (approved):
Awesome 👍

PR #33, opened 2022-01-08
Fix erroneously returning "success" when the command failed.

RenierM26:
* Fix switchbot erroneously returning "success" on execution when signal is weak/intermittent. 
* Workaround for erroneously returning "failed" when a disconnect occurs after the command is sent. (Read notification step)


Danielhiversen (approved):
👍

PR #40, opened 2022-02-19
Timeout when unable to connect to device after 40 seconds.

RenierM26:
- Improve handling of disc state.
- Default Connection timeout, instead of bluepy default, if unable to connect to device after 40 seconds.

### DIFF
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -12,7 +12,6 @@ import bluepy
 DEFAULT_RETRY_COUNT = 3
 DEFAULT_RETRY_TIMEOUT = 1
 DEFAULT_SCAN_TIMEOUT = 5
-Debugging = False
 
 # Keys common to all device types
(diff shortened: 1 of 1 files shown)


pascalwinters:
_[copied to related PR]_
@RenierM26 For me it's stable since the beginning of using this fix. Two remarks/questions:

1. Maybe it's good (before merge this PR), to mark this line as DEBUG:
2022-02-20 11:12:40 WARNING (SyncWorker_2) [switchbot] Bluepy busy, waiting before retry

2. I'm missing the log of reading each (I mean) 60 seconds the current status. Is this log event deleted?

3. Is it an idea to merge this and make a PR for HA (I can do that if you want) so it has been fixed for all the users in HA with the next release. But we must make progress, because the beta is almost there.

pascalwinters:
> @RenierM26 @pascalwinters as this has already been merged, do we need to open a PR for HA?

No, it’s aleady done and merged 😉
https://github.com/home-assistant/core/pull/67025

PR #43, opened 2022-06-06
[Patch] _readkey only reads if BLE device supports reading & corrected Exception when reading

ninadpchaudhari:
This is _not_ a breaking change, can be released before the shift away from bluepy. 
- Solves the issue of `Bluetooth command failed (code: 2, error: Attribute can't be read)` in Switchbot Bots.
- Corrected `BTLEException` to `BTLEGattError`, as that is thrown for `read` function.
- Read operation will only be executed if the device supports reading. for eg. curtain


ninadpchaudhari:
Closing this for Made a better PR #44

PR #44, opened 2022-06-08
[Patch] _readkey only reads if BLE device supports reading & corrected Exception when reading

ninadpchaudhari:
This is _not_ a breaking change, can be released before the shift away from bluepy. 
- Solves the issue of `Bluetooth command failed (code: 2, error: Attribute can't be read)` in Switchbot Bots.
- Corrected `BTLEException` to `BTLEGattError`, as that is thrown for `read` function.
- Read operation will only be executed if the device supports reading. for eg. curtain

### DIFF
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -372,25 +372,32 @@ class SwitchbotDevice(bluepy.btle.Peripheral):
         return True
 
     def _readkey(self) -> bytes:
+        """Reads Bytes from BLE device.
+        
+        Returns:
+            bytes: Response from BLE device or b'\00' otherwise.
+        """
         _LOGGER.debug("Prepare to read notification from switchbot")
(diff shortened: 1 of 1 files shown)


RenierM26:
Hi @ninadpchaudhari,

The correct way is actually to parse notifications instead of reading a characteristic. Switchbot devices "respond" to any command issued.

https://ianharvey.github.io/bluepy-doc/notifications.html

It will result in one less command to the switchbot that could fail. 

These devices just don't handle communication reliably so less commands would improve reliability. (I have a suspicion it might be due to negotiated protocol settings or something but still need to find a way to increase the actual connection timeout - I think the default for bluez is just too short)

bdraco:
Looks like this PR has gone stale. If there is still an issue, please open a fresh PR.

Thanks @ninadpchaudhari

PR #49, opened 2022-07-23
Refactor to support push updates

bdraco:
- Fixes the assumption that address were mac addresses
  on MacOS they are UUIDs

- Adds the ability to consume an update from the scanner
  that is always running


bdraco:
Tested with bots, curtain, and temp sensors
https://github.com/home-assistant/core/pull/75645

Danielhiversen (approved):
Awesome. Thanks

bdraco:
Thanks

PR #48, opened 2022-07-16
Remove filter to handle new UUIDs in SwitchBot firmware

Gibstick:
The new SwitchBot firmware advertises its services differently, making the old UUID filter not find any SwitchBot devices. This change is a quick-fix to remove the filter and simply allow the rest of the discovery code to handle it as before.

I don't really know Bluetooth so this is just a total hack that seemed to work. But from reading of the rest of the code, it's still handling advertisement data and making sure it's a supported model, so this change shouldn't hurt. 

In theory, this should be backwards compatible with the old firmware since the rest of the discovery and initialization is exactly the same, but I don't have many devices to test with. Running a scan, I was able to find my own SwitchBot thermometer, and a neighbour's "WoHand" device, which I'm assuming is a switch. Before, I wasn't able to find any devices.

See also https://github.com/devWaves/SwitchBot-MQTT-BLE-ESP32/issues/80

This seems to address https://github.com/home-assistant/core/issues/70262, since after this patch I was able to have HomeAssistant discover my neighbour's switch.

PR #50, opened 2022-07-24
Use bleak-retry-connector to handle transient connection errors

bdraco:
We built bleak-retry-connector for aiohomekit to handle the transient errors that happen when using bleak with dbus since
we were missing disconnected events. As switchbots seem to fail in the same way with bleak on dbus, using the library
made the problem go away.

Also, the default connect timeout is too low which means [services cannot be resolved in time](https://github.com/hbldh/bleak/blob/0361ba4b7e4682f4e97ecacaab4499b84c70f288/bleak/backends/bluezdbus/client.py#L316) so it disconnects in the middle of the connection leaving things in a bad state which is likely why the retry timeout option was added.  Although it might have been added for not finding the device via discovery and then added to the client path as well.  It makes sense in the scanner path since sometimes the device can fall out of the list, but shouldn't be needed in the client path as long as the connect timeout doesn't cause a disconnect in the middle of services being resolved which the current low timeout will do.

The short version is we don't count transient errors as needing to wait, and since they seem to happen often, it makes the action quite a bit faster if we hit a transient one as we try again right away instead of sleeping.

This should also fix the issue reported here https://github.com/Danielhiversen/pySwitchbot/pull/48#issuecomment-1193113155

```
2022-07-23 21:53:42.549 DEBUG (MainThread) [bleak_retry_connector] FD:60:30:55:92:57: WoHand: Connecting (attempt: 1)
2022-07-23 21:53:44.820 DEBUG (MainThread) [bleak_retry_connector] FD:60:30:55:92:57: WoHand: Failed to connect: [org.bluez.Error.Failed] le-connection-abort-by-local (attempt: 1)
2022-07-23 21:53:44.820 DEBUG (MainThread) [bleak_retry_connector] FD:60:30:55:92:57: WoHand: Connecting (attempt: 2)
2022-07-23 21:53:45.843 DEBUG (MainThread) [bleak_retry_connector] FD:60:30:55:92:57: WoHand: Failed to connect: [org.bluez.Error.Failed] le-connection-abort-by-local (attempt: 2)
2022-07-23 21:53:45.843 DEBUG (MainThread) [bleak_retry_connector] FD:60:30:55:92:57: WoHand: Connecting (attempt: 3)
2022-07-23 21:53:47.544 DEBUG (MainThread) [bleak_retry_connector] FD:60:30:55:92:57: WoHand: Connected (attempt: 3)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup
 setup(
     name = 'PySwitchbot',
     packages = ['switchbot'],
-    install_requires=['bleak'],
+    install_requires=['bleak', 'bleak-retry-connector'],
     version = '0.15.0',
     description = 'A library to communicate with Switchbot',
     author='Daniel Hjelseth Hoyer',
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -11,6 +11,7 @@ from uuid import UUID
 import bleak
 from bleak.backends.device import BLEDevice
 from bleak.backends.scanner import AdvertisementData
+from bleak_retry_connector import BleakClient, establish_connection
 
 DEFAULT_RETRY_COUNT = 3
 DEFAULT_RETRY_TIMEOUT = 1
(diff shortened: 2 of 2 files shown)


Danielhiversen (approved):

bdraco:
Thanks 👍

PR #51, opened 2022-07-25
Add support for contact sensors

bdraco:
I wanted to have another device to test with for the push updates so I ordered the contact sensors and the motion sensor.  The contact sensor showed up first so I'm adding that first.

I'll do a PR for the motion sensor when it shows.

This also adds a mapping for the `plus` meter models which are the same as the standard meters with a larger screen


bdraco:
@Danielhiversen if you get a chance, it would be great to get this one and https://github.com/Danielhiversen/pySwitchbot/pull/52 in beta

Danielhiversen (approved):

bdraco:
Thanks

PR #52, opened 2022-07-26
Use an asyncio.Future for commands instead of a sleep

bdraco:
Speeds up sending commands

```
2022-07-25 22:16:23.775 DEBUG (MainThread) [switchbot] Subscribe to notifications
2022-07-25 22:16:23.826 DEBUG (MainThread) [switchbot] Sending command, 570102
2022-07-25 22:16:23.858 INFO (MainThread) [switchbot] Notification received: bytearray(b'\x05\xff\x00')
2022-07-25 22:16:23.858 DEBUG (MainThread) [switchbot] UnSubscribe to notifications
```

PR #53, opened 2022-07-26
Add basic support for the mini plugs

bdraco:


bdraco:
@Danielhiversen the motion sensor and the lights I ordered for testing showed up so I have a motion sensor PR almost ready to go.

It would be great if you could merge this one and https://github.com/Danielhiversen/pySwitchbot/pull/52 and then I'll start the motion sensor PR since it will conflict.

Thanks 
👍

PR #54, opened 2022-07-29
Fix connecting with bleak 0.15

bdraco:
This will work with 0.14.3 as well

There is also a performance fix in #54

I merged the performance fix from #52 into #54 and closed #52 since it would conflict

### DIFF
--- a/switchbot/__init__.py
+++ b/switchbot/__init__.py
@@ -325,23 +325,29 @@ class SwitchbotDevice:
 
         raise RuntimeError("Unreachable")
 
+    @property
+    def name(self) -> str:
+        """Return device name."""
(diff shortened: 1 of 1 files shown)


pascalwinters:
@Danielhiversen can you merge this PR, so we can test it in the latest HA beta? The SwitchBot integration is broken now because of the Bleak version.

DarKOrange75:
Integration broken for me after upgrade to 2022.8.0 public release :-(

bdraco:
`org.bluez.Device1  doesn't exist` means the bluetooth adapter can't be found.  

Try restarting the host and make sure `bluetoothctl` can see the adapter.

This isn't the place to get bluetooth support though

PR #55, opened 2022-07-29
Add support for the motion sensor

bdraco:
Needs https://github.com/Danielhiversen/pySwitchbot/pull/53 first


bdraco:
Actual naming should be WoPresence

PR #74, opened 2022-08-14
Add support for the bulbs

bdraco:


bdraco:
We need to poll when the counter is incremented for the extended data not in the adv

bdraco:
But it needs a debounce

bdraco:
Will wait for similar efforts on debounce in https://github.com/Jc2k/aiohomekit/pull/151 to move this forward

PR #127, opened 2022-10-07
Send both curtain operation commands

bdraco:
Since we do not know which command set the device needs we send both


bdraco:
I ended up purchasing a second switchbot curtain to test this and verified it fixed the issue reported in https://github.com/home-assistant/core/issues/79827

PR #128, opened 2022-10-10
Add paired field

bdraco:
I think this is wrong actually.  It might be right/left but its not documented so need to do some more experiments


bdraco:
It does appear if paired = True its the child and controlling the other one will affect it

PR #145, opened 2022-12-04
Add read-only support for lock

dsypniewski:
Based on recently released but not yet merged documentation for the lock (https://github.com/OpenWonderLabs/SwitchBotAPI-BLE/blob/4261be8ec49f8b181c54b786b3479c03ad5f80b8/devicetypes/lock.md)
Tested on a sample of one, so I couldn't verify the data related to double lock mode but maybe someone else can verify that part.
Let's hope SwitchBot releases actions for the lock sooner rather than later, but in the meantime maybe we could at least enjoy the read-only part.

PS. after reading through code I wasn't sure which naming convention should I use do I went with snake_case but I can change that if need be


bdraco (approved):

bdraco:
Thanks @dsypniewski

PR #162, opened 2022-12-18
Always fire callbacks when we do a full update since all sensors will update

bdraco:


switchbot/devices/device.py
```diff
@@ -579,6 +579,12 @@ def switch_mode(self) -> bool | None:
 
     def poll_needed(self, seconds_since_last_poll: float | None) -> bool:
         """Return if device needs polling."""
+        _LOGGER.warning(
+            "%s: Polling needed: seconds_since_last_poll=%s, _last_full_update=%s",
+            self.name,
+            seconds_since_last_poll,
+            self._last_full_update,
+        )
```
bdraco, line 587:
```suggestion
```
left in debug

PR #164, opened 2022-12-23
Added lock model with lock and unlock support

dsypniewski:
As stated in the title, after reverse engineering the communication between lock and the app here it is.

Script is meant to retrieve key from the internal switchbot app API, and values obtained from that API have to be used as constructor parameters for the model.
This also means that the lock first has to be setup with the app before obtaining that key.
Script uses boto3 as a dependency because switchbot is using AWS as an authentication provider, but I haven't added that to library dependencies because it's not needed after getting the key.
In the script there are hard coded AWS resource IDs which were obtained from resource files extracted from the Android app, there is a comment in there so those value can be relatively easily verified by anyone with apk file of the app.


setup.py
```diff
@@ -3,7 +3,7 @@
 setup(
     name="PySwitchbot",
     packages=["switchbot", "switchbot.devices", "switchbot.adv_parsers"],
-    install_requires=["async_timeout>=4.0.1", "bleak>=0.17.0", "bleak-retry-connector>=2.9.0"],
+    install_requires=["async_timeout>=4.0.1", "bleak>=0.17.0", "bleak-retry-connector>=2.9.0", "pycryptodome>=3.16.0"],
```
bdraco, line 6: Home assistant uses the cryptography instead. Can you switch this out?

dsypniewski: Done

switchbot/devices/lock.py
```diff
@@ -1 +1,120 @@
...
+
+    def __init__(self, device: BLEDevice, key_id: str, encryption_key: str, interface: int = 0, **kwargs: Any) -> None:
+        if len(key_id) == 0:
+            _LOGGER.error("key_id is missing")
+        elif len(key_id) != 2:
+            _LOGGER.error("invalid key_id")
+        if len(encryption_key) == 0:
+            _LOGGER.error("encryption_key is missing")
+        elif len(encryption_key) != 32:
+            _LOGGER.error("invalid encryption_key")
```
bdraco, line 32: These should probably raise `ValueError` instead.

switchbot/devices/lock.py
```diff
@@ -1 +1,120 @@
...
+        self._key_id = key_id
+        self._encryption_key = bytearray.fromhex(encryption_key)
+        super().__init__(device, None, interface, **kwargs)
+
+    async def lock(self) -> bool:
+        """Send lock command."""
+        result = await self._send_command(COMMAND_LOCK)
+        return self._check_command_result(result, 0, {1})
+
+    async def unlock(self) -> bool:
```
bdraco, line 44: https://github.com/Danielhiversen/pySwitchbot/blob/a804c7863b253dfe0b09bdca3cd724d21c201492/switchbot/devices/curtain.py#L80

Probably should decorate this with `@update_after_operation` to make sure the locking worked

switchbot/devices/lock.py
```diff
@@ -1 +1,120 @@
...
+            _LOGGER.error("encryption_key is missing")
+        elif len(encryption_key) != 32:
+            _LOGGER.error("invalid encryption_key")
+        self._iv = None
+        self._cipher = None
+        self._key_id = key_id
+        self._encryption_key = bytearray.fromhex(encryption_key)
+        super().__init__(device, None, interface, **kwargs)
+
+    async def lock(self) -> bool:
```
bdraco, line 39: https://github.com/Danielhiversen/pySwitchbot/blob/a804c7863b253dfe0b09bdca3cd724d21c201492/switchbot/devices/curtain.py#L80

Probably should decorate this with `@update_after_operation` to make sure the locking worked

switchbot/devices/lock.py
```diff
@@ -1 +1,120 @@
...
+        result = await self._send_command(COMMAND_LOCK)
+        return self._check_command_result(result, 0, {1})
+
+    async def unlock(self) -> bool:
+        """Send unlock command."""
+        result = await self._send_command(COMMAND_UNLOCK)
+        return self._check_command_result(result, 0, {1})
+
+    async def update(self, interface: int | None = None) -> None:
+        await self.get_device_data(retry=self._retry_count, interface=interface)
```
bdraco, line 50: The implementation of `update` has changed in `device.py` to work with passive advertisements. I think you can drop `update` here and let it use the parent class

switchbot/devices/lock.py
dsypniewski: I tried adding that but it seems that it doesn't do anything, the lock takes a bit of time to change state so immediate update after the action doesn't catch the state change, hmm...

bdraco: You'll have to implement `get_basic_info` first

dsypniewski: Yes I did that but the status doesn't update anyway, as I said I think it's because the locking command sends response right away and doesn't wait for locking to be finished. If I wait a few second after executing this method and then do update it works.
So it takes a few seconds after the current lock method (together with the update after) is finished to update the actual status in the advertisements.
Should I add something in those methods to wait for the actual status change?

bdraco: You could do a call_later with a debounce to poll it a bit later

bdraco: Take a look at the timed disconnect code in parent class. You can probably borrow most of that

switchbot/devices/lock.py
```diff
@@ -1 +1,157 @@
...
+        await self.update()
+        status = self.get_lock_status()
+        if status == self._status_before_update or status in {LockStatus.LOCKING, LockStatus.UNLOCKING}:
+            self._set_status_update_timer(status)
+        else:
+            self._status_before_update = None
+
+    async def get_basic_info(self) -> dict[str, Any] | None:
+        """Get device basic settings."""
+        return (await self.get_device_data()).data["data"]
```
bdraco, line 87: Is there a way you poll the characteristic for status instead since you are actively connected? The advertisement data tends to be stale when connected to the device on the other SwitchBot devices

bdraco:
I ordered one of these locks so I can test this

switchbot/devices/lock.py
dsypniewski: That's what I'm seeing that it takes quite a bit of time to update - sometimes 2 seconds and sometimes more like 10, but unfortunately I haven't yet seen lock status (and other details from advertisement) in responses from characteristic.

bdraco: If you send https://github.com/Danielhiversen/pySwitchbot/blob/a804c7863b253dfe0b09bdca3cd724d21c201492/switchbot/devices/device.py#L36 do you get back state?

dsypniewski: I get a response back but as far as I can tell it only contains better percentage, firmware version, one more value I haven't been able to correlate yet and some 0s. I've tried executing this in different states of the lock but it doesn't change at all.

dsypniewski:
I found another solution to getting the state of the lock, so I'll update this PR once I have that implementation done.

PR #167, opened 2022-12-27
Changed lock model constructor to not require advertisement data

dsypniewski:
On HA side the advertisement data is actually not available when the device model instance gets created.
This PR updates the Lock model to be more in line with other device models as it should have been in the original PR #164


bdraco (approved):

bdraco:
Released in 0.32.0

I didn't bump Home Assistant so you'll have to do that in your PR there.

PR #170, opened 2022-12-29
Add retrieve_encryption_key for lock

dsypniewski:
In order to address this comment https://github.com/home-assistant/core/pull/84673#issuecomment-1366562534 I've moved the encryption key retrieval code into the library itself.


bdraco:
Ideally this would use asyncio but that's not too big of a concern since it's expected to be called rarely

bdraco:
0.34.0

PR #181, opened 2023-01-07
Initial Blind Tilt support

### LINKED ISSUE #172: Adding support for new SwitchBot Blind Tilt
Hi,

I just received my [SwitchBot Blind Tilt](https://www.kickstarter.com/projects/switchbot/switchbot-blind-tilt-solar-powered-retrofit-smart-blinds) , and I would like to see if I can be of any assistance to help adding support to pySwitchbot. For example,  is there any bluetooth capture I could provide to help more capable people implement the protocol for these new devices ?

Regards

jesserockz:
closes #172

### DIFF
--- a/switchbot/adv_parsers/blind_tilt.py
+++ b/switchbot/adv_parsers/blind_tilt.py
@@ -12,7 +12,7 @@ def process_woblindtilt(
 
     device_data = mfr_data[6:]
 
-    _tilt = max(min(device_data[2] & 0b00111111, 100), 0)
+    _tilt = max(min(device_data[2] & 0b01111111, 100), 0)
     _in_motion = bool(device_data[2] & 0b10000000)
     _light_level = (device_data[1] >> 4) & 0b00001111
 
--- a/switchbot/adv_parsers/blind_tilt.py
+++ b/switchbot/adv_parsers/blind_tilt.py
@@ -1,6 +1,7 @@
 """Library to handle connection with Switchbot."""
 from __future__ import annotations
 
+
 def process_woblindtilt(
     data: bytes | None, mfr_data: bytes | None, reverse: bool = False
 ) -> dict[str, bool | int]:
--- a/switchbot/adv_parsers/blind_tilt.py
+++ b/switchbot/adv_parsers/blind_tilt.py
@@ -7,6 +7,9 @@ def process_woblindtilt(
 ) -> dict[str, bool | int]:
     """Process woBlindTilt services data."""
 
+    if mfr_data is None:
+        return {}
+
     device_data = mfr_data[6:]
 
     _tilt = max(min(device_data[2] & 0b00111111, 100), 0)
--- a/switchbot/adv_parsers/blind_tilt.py
+++ b/switchbot/adv_parsers/blind_tilt.py
@@ -1,9 +1,8 @@
 """Library to handle connection with Switchbot."""
 from __future__ import annotations
 
-
 def process_woblindtilt(
-    data: bytes | None, mfr_data: bytes | None, reverse: bool = True
+    data: bytes | None, mfr_data: bytes | None, reverse: bool = False
 ) -> dict[str, bool | int]:
     """Process woBlindTilt services data."""
 
@@ -15,9 +14,10 @@ def process_woblindtilt(
     _tilt = max(min(device_data[2] & 0b01111111, 100), 0)
     _in_motion = bool(device_data[2] & 0b10000000)
     _light_level = (device_data[1] >> 4) & 0b00001111
+    _calibrated = bool(device_data[1] & 0b00000001)
 
     return {
-        "calibration": bool(device_data[1] & 0b00000001),
+        "calibration": _calibrated,
         "battery": data[2] & 0b01111111 if data else None,
         "inMotion": _in_motion,
         "tilt": (100 - _tilt) if reverse else _tilt,
--- a/switchbot/adv_parsers/blind_tilt.py
+++ b/switchbot/adv_parsers/blind_tilt.py
@@ -1,0 +1,22 @@
+"""Library to handle connection with Switchbot."""
+from __future__ import annotations
+
+
+def process_woblindtilt(
+    data: bytes | None, mfr_data: bytes | None, reverse: bool = True
+) -> dict[str, bool | int]:
+    """Process woBlindTilt services data."""
+
+    device_data = mfr_data[6:]
+
+    _tilt = max(min(device_data[2] & 0b00111111, 100), 0)
+    _in_motion = bool(device_data[2] & 0b10000000)
+    _light_level = (device_data[1] >> 4) & 0b00001111
+
+    return {
+        "calibration": bool(data[1] & 0b01000000) if data else None,
+        "battery": data[2] & 0b01111111 if data else None,
+        "inMotion": _in_motion,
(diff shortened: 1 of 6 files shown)


switchbot/adv_parsers/blind_tilt.py
```diff
@@ -0,0 +1,22 @@
+"""Library to handle connection with Switchbot."""
+from __future__ import annotations
+
+
+def process_woblindtilt(
+    data: bytes | None, mfr_data: bytes | None, reverse: bool = True
+) -> dict[str, bool | int]:
+    """Process woBlindTilt services data."""
+
```
bdraco, line 9: `mfr_data ` can be likely be `None` if its anything like the other devices so it needs a guard here to return `None`

switchbot/adv_parsers/blind_tilt.py
```diff
@@ -0,0 +1,25 @@
...
+    if mfr_data is None:
+        return {}
+
+    device_data = mfr_data[6:]
+
+    _tilt = max(min(device_data[2] & 0b00111111, 100), 0)
```
BTMorton, line 15: I think this should be:
```suggestion
    _tilt = max(min(device_data[2] & 0b01111111, 100), 0)
```
Otherwise it limits the tilt value to 0-63

jesserockz: This could explain some issues I was having =)

switchbot/devices/blind_tilt.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class SwitchbotBlindTilt(SwitchbotCurtain):
+    """Representation of a Switchbot Blind Tilt."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Switchbot Blind Tilt/woBlindTilt constructor."""
+
+        super().__init__(*args, **kwargs)
+
```
bdraco, line 19:
```suggestion
```

Looks like a useless constructor?

BTMorton:
I'm wondering if it's worth doing `open`, `close_up`, and `close_down` methods for the blind tilts like offered in the app, rather than just `open`/`close` like the curtains. Not 100% sure how Home Assistant would handle it. Might result in only supporting open and one of the two close directions, but it would better reflect the unit's actual behaviour. Happy to do the impl?

Also, should there be another helper method in `discovery.py`?
```
    async def get_blind_tilts(self) -> dict[str, SwitchBotAdvertisement]:
        """Return all WoBlindTilt/BlindTilts devices with services data."""
        return await self._get_devices_by_model("x")
```

switchbot/devices/blind_tilt.py
```diff
@@ -0,0 +1,59 @@
...
+import logging
+from typing import Any
+
+from .curtain import SwitchbotCurtain, CURTAIN_EXT_SUM_KEY
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class SwitchbotBlindTilt(SwitchbotCurtain):
+    """Representation of a Switchbot Blind Tilt."""
```
BTMorton, line 13: Here's a proposal implementation to add open/close_up/close_down methods. I've added a close method that will finish closing the blinds in the direction they are already partially open, defaulting to close down when fully open (50%).

```suggestion
BLIND_COMMAND = "4501"
OPEN_KEYS = [
    f"{REQ_HEADER}{BLIND_COMMAND}010132",
    f"{REQ_HEADER}{BLIND_COMMAND}05ff32",
]
CLOSE_DOWN_KEYS = [
    f"{REQ_HEADER}{BLIND_COMMAND}010100",
    f"{REQ_HEADER}{BLIND_COMMAND}05ff00",
]
CLOSE_UP_KEYS = [
    f"{REQ_HEADER}{BLIND_COMMAND}010164",
    f"{REQ_HEADER}{BLIND_COMMAND}05ff64",
]

class SwitchbotBlindTilt(SwitchbotCurtain):
    """Representation of a Switchbot Blind Tilt."""

    @update_after_operation
    async def open(self) -> bool:
        """Send open command."""
        return await self._send_multiple_commands(OPEN_KEYS)

    @update_after_operation
    async def close_up(self) -> bool:
        """Send close up command."""
        return await self._send_multiple_commands(CLOSE_UP_KEYS)

    @update_after_operation
    async def close_down(self) -> bool:
        """Send close down command."""
        return await self._send_multiple_commands(CLOSE_DOWN_KEYS)

    @update_after_operation
    async def close(self) -> bool:
        """Send close command."""
        if (self.get_position() > 50):
            return await self.close_up()
        else:
            return await self.close_down()
```

Also needs the following imports:
```
from switchbot.devices.device import REQ_HEADER, update_after_operation
```

switchbot/adv_parsers/blind_tilt.py
```diff
@@ -0,0 +1,25 @@
...
+        return {}
+
+    device_data = mfr_data[6:]
+
+    _tilt = max(min(device_data[2] & 0b01111111, 100), 0)
+    _in_motion = bool(device_data[2] & 0b10000000)
+    _light_level = (device_data[1] >> 4) & 0b00001111
+
+    return {
+        "calibration": bool(data[1] & 0b01000000) if data else None,
```
BTMorton, line 20: I don't know that this is the calibration flag. This byte is always 0 on my devices. I'm going to investigation further to see if I can figure out if there is another bit being used.

BTMorton: Figured it out. It's the first bit of `device_data[1]`

```suggestion
        "calibration": bool(device_data[1] & 0b00000001),
```

BTMorton:
I did the changes I proposed to add `close_up`/`close_down` methods in a new branch, and did an integration with Home Assistant. This is tested and running with the blind tilt devices in my local install.

Lib changes: https://github.com/Danielhiversen/pySwitchbot/compare/master...BTMorton:pySwitchbot:add_blind_tilt_support
HA integration: https://github.com/home-assistant/core/compare/dev...BTMorton:home-assistant-core:add-switchbot-blind-tilt-support

It works pretty nicely as an integration as it matches the up/down arrows in the UI. The problem is that the behaviour isn't super intuitive from the cover service naming. Calling "open" when the blind is already open will close the blind up, because the "open" arrow in the UI points up. Likewise when it's closed, calling "close" will open the blind. It makes sense from an interface perspective, but not so much on the automation front.

There's potentially a limitation in the base cover tilt implementation in that it doesn't seem to support bi-directional closing. Making an assumption and limiting it to close one-way or the other will upset someone somewhere, although it could be worked around by calibrating the device in reverse.

This solution was a "light-touch" way to add full support. I'll leave it to you guys to decide next steps.

If this is the wrong place to put this, let me know and I can re-post somewhere more appropriate.

bdraco:
@BTMorton maybe send a PR to Jesse's branch?

jesserockz:
Please feel free to PR into this branch. I have had no time myself to look into this so I am glad someone is.

In regards to the HA side, I did implement it a little, but because this is setting the tilt of a curtain/blind, I made the entity only support tilt, not actual position with open/close functions which better represents what it actually is. Feel free to disagree there.

jesserockz:
With the suggestions by @BTMorton merged, this now opens and closes nicely from HA. The default HA frontend controls don't like to allow `open` when the position is already at 100, but the service calls to open work, and it opens to 50% nicely.

PR #182, opened 2023-01-07
Allow low battery status responses for lock

dsypniewski:
This should fix a bug where responses have different status code when lock has low battery.


bdraco:
Thanks @dsypniewski

JamieSBenson:
@dsypniewski @bdraco Awesome, thank you folks for all you do!

PR #184, opened 2023-01-11
Fix contact_open timeout

dcousens:
(see https://github.com/OpenWonderLabs/SwitchBotAPI-BLE/issues/15#issuecomment-1368175346)

As is, the contact only "stays" open for 60 seconds, which is wrong.

### DIFF
--- a/switchbot/adv_parsers/contact.py
+++ b/switchbot/adv_parsers/contact.py
@@ -15,7 +15,7 @@ def process_wocontact(
 
     if mfr_data and len(mfr_data) >= 13:
         motion_detected = bool(mfr_data[7] & 0b10000000)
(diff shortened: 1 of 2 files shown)


bdraco:
@dcousens I tweaked it a bit and fixed the test.  Does this still work for you as expected?

dcousens:
@bdraco I don't use this library, unfortunately. https://github.com/Danielhiversen/pySwitchbot/pull/184/commits/7c62122b93109df70ebefa40dc382db59d94b0cc LGTM though

bdraco:
Looks like I need to dig this sensor back out of storage for testing 🤣

bdraco:
confirmed the fix when active scans are disabled 👍

PR #215, opened 2023-08-16
Added support for Night Latch feature of EU firmware version of smart lock

Michal4K:
Added support for Night Latch feature of EU firmware version of smart lock

Adding:
- detection if Night Latch feature is enabled from ADV
- new lock function to unlatch
- Changing way of handling unlock command. Night Latch off -> same - sends unlock command, Night Latch on -> sends command not fully unlock, because unlock command in this configuration also unlatches door

tested with firmware version 6.5


switchbot/devices/lock.py
```diff
@@ -165,6 +166,16 @@ async def lock(self) -> bool:
...
+            )
+        return await self._lock_unlock(
+            COMMAND_UNLOCK, {LockStatus.UNLOCKED, LockStatus.UNLOCKING, LockStatus.NOT_FULLY_LOCKED}
+        )
+
+    async def unlatch(self) -> bool:
+        """Send unlatch command. Unlatches door if feature is enabled elsewhere just unlock them."""
         return await self._lock_unlock(
             COMMAND_UNLOCK, {LockStatus.UNLOCKED, LockStatus.UNLOCKING}
         )
```
bdraco, line 181: Please make separate functions for each command.  We don't want conditionals in the functions.  The consumer of the api should call the correct one based on what they want, no what the lib thinks they want

switchbot/devices/lock.py
```diff
@@ -26,6 +26,7 @@
 COMMAND_GET_CK_IV = f"{COMMAND_HEADER}0f2103"
 COMMAND_LOCK_INFO = f"{COMMAND_HEADER}0f4f8101"
 COMMAND_UNLOCK = f"{COMMAND_HEADER}0f4e01011080"
+COMMAND_NOT_FULLY_UNLOCK = f"{COMMAND_HEADER}0f4e010110a0"
```
bdraco, line 29: This seems like it should be called unlatch. Not fully unlock doesn't really explain what's going on here.

I wouldn't expect someone to ask me to not fully unlock a door

bdraco: Another way to think about it would be: what would you ask a voice assistant to do to execute this action

Michal4K: I am not native English speaker. But it seems weird to me call this command unlatch if this just unlock door... not hold the latch. I don't know why they implemented it this way. Unlock command holds the latch if feature is enabled. It would make more sense if they make new command for unlatching not for unlocking without holding the latch.

bdraco: I'm not 100% sure what the new command does as I don't have the lock, but maybe `COMMAND_UNLOCK_WITHOUT_UNLATCH`

Michal4K: Renaming unlock to open and not_fully_unlock to unlock would be nice. But it would break backwards compatibility so... its not probably the best idea.

bdraco: maybe add `open` and `unlock_and_open`

Michal4K: I don't know about that. That seems misleading too to me. The name NOT_FULLY_UNLOCK is at least some how consistent with name of the state that the lock return NOT_FULLY_LOCKED.

Michal4K:
Sorry, for not running tests before opening PR.

switchbot/devices/lock.py
bdraco: Lets call it `unlock_without_unlatch`

bdraco (approved):

PR #217, opened 2023-08-29
Fix compatibilty for old/US firmware

Michal4K:
Fixing compatibility with older lock firmware / US firmware. I don't know which versions exactly have this problem - fixing it to address (https://github.com/home-assistant/core/pull/98620#pullrequestreview-1600316671)

- added test cases for one byte shorter manufacturer_data


bdraco:
[0.39.1](https://github.com/Danielhiversen/pySwitchbot/releases/tag/0.39.1) published

PR #220, opened 2023-09-30
Fix off-by-one error in curtain parser

the-ress:
Fixes https://github.com/home-assistant/core/issues/101157.

Line 12 probably should be changed to `>= 12` too, but I wanted to keep the changes to a minimum so I don’t risk breaking anything else for HA 2023.10.


bdraco (approved):

bdraco:
0.40.1

PR #221, opened 2023-11-11
Ceiling light pro

Plemeur:
Hello,
This add support for the light functionality of the Switchbot Ceiling Light pro.

Tested in home assistant, the light is detected and the control are working as expected

There is a device with model "m" as well when I run the device discovery, this might be the hub inside the lamp


bdraco (approved):

bdraco:
Thanks @Plemeur

PR #226, opened 2024-01-10
Add support for is_opening and is_closing for curtains

dcmeglio:
This is to support a fix for Home-Assistant for https://github.com/home-assistant/core/issues/99753

Currently no information is provided to Home-Assistant or any consumer to provide information about the direction of travel. The message has an inMotion value, but this is just boolean. This adds `is_opening()` and `is_closing()` values to allow Home-Assistant (and other consumers) to know whether the curtain is opening or closing. This is done by comparing the current position to the previous position to know which direction the device moves.

I have tested with both curtains that open to the left and the right but I do not have one that opens from the center (paired) to test with. However, I also added several new unit tests (needed to add a dependency on pytest-mock) to test as many scenarios as possible in addition to the existing advertisement parser tests.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,6 +1,5 @@
 pytest-asyncio
 pytest-cov
-pytest-mock
 bleak>=0.17.0
 bleak-retry-connector>=3.4.0
 cryptography>=38.0.3
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,5 +1,6 @@
 pytest-asyncio
 pytest-cov
+pytest-mock
 bleak>=0.17.0
 bleak-retry-connector>=3.4.0
 cryptography>=38.0.3
--- a/switchbot/devices/curtain.py
+++ b/switchbot/devices/curtain.py
@@ -249,11 +249,10 @@ class SwitchbotCurtain(SwitchbotDevice):
         # To get actual light level call update() first.
         return self._get_adv_value("calibration")
 
-    @property
-    def is_opening(self) -> Any:
+    def is_opening(self) -> bool:
         """Return True if the curtain is opening."""
         return self._is_opening
-    @property
-    def is_closing(self) -> Any:
+    
+    def is_closing(self) -> bool:
         """Return True if the curtain is closing."""
         return self._is_closing
--- a/switchbot/devices/curtain.py
+++ b/switchbot/devices/curtain.py
@@ -1,8 +1,11 @@
 """Library to handle connection with Switchbot."""
 from __future__ import annotations
+from dataclasses import replace
 
 import logging
 from typing import Any
+
+from switchbot.models import SwitchBotAdvertisement
 
 from .device import REQ_HEADER, SwitchbotDevice, update_after_operation
 
@@ -54,6 +57,22 @@ class SwitchbotCurtain(SwitchbotDevice):
         self._settings: dict[str, Any] = {}
         self.ext_info_sum: dict[str, Any] = {}
         self.ext_info_adv: dict[str, Any] = {}
+        self._is_opening: bool = False
+        self._is_closing: bool = False
+
+    def _set_parsed_data(
+        self, advertisement: SwitchBotAdvertisement, data: dict[str, Any]
+    ) -> None:
+        """Set data."""
+        _new_data = replace(
+            advertisement, data=self._sb_adv_data.data | {"data": data}
+        )
+        self._update_motion_direction(_new_data["inMotion"], self._get_adv_value("position"), _new_data["position"])
+        if _new_data["inMotion"] == True:
+            pass
+        else:
+            self._is_closing = self._is_opening = False
+        super()._set_parsed_data(advertisement, data)
 
     async def _send_multiple_commands(self, keys: list[str]) -> bool:
         """Send multiple commands to device.
@@ -70,6 +89,8 @@ class SwitchbotCurtain(SwitchbotDevice):
     @update_after_operation
     async def open(self, speed: int = 255) -> bool:
         """Send open command. Speed 255 - normal, 1 - slow"""
+        self._is_opening = True
+        self._is_closing = False
         return await self._send_multiple_commands(
             [OPEN_KEYS[0], f"{OPEN_KEYS[1]}{speed:02X}00"]
         )
@@ -77,6 +98,8 @@ class SwitchbotCurtain(SwitchbotDevice):
     @update_after_operation
     async def close(self, speed: int = 255) -> bool:
         """Send close command. Speed 255 - normal, 1 - slow"""
+        self._is_closing = True
+        self._is_opening = False
         return await self._send_multiple_commands(
             [CLOSE_KEYS[0], f"{CLOSE_KEYS[1]}{speed:02X}64"]
         )
@@ -84,12 +107,14 @@ class SwitchbotCurtain(SwitchbotDevice):
     @update_after_operation
     async def stop(self) -> bool:
         """Send stop command to device."""
+        self._is_opening = self._is_closing = False
         return await self._send_multiple_commands(STOP_KEYS)
 
     @update_after_operation
     async def set_position(self, position: int, speed: int = 255) -> bool:
         """Send position command (0-100) to device. Speed 255 - normal, 1 - slow"""
         position = (100 - position) if self._reverse else position
+        self._update_motion_direction(True, self._get_adv_value("position"), position)
         return await self._send_multiple_commands(
             [
                 f"{POSITION_KEYS[0]}{position:02X}",
@@ -108,6 +133,11 @@ class SwitchbotCurtain(SwitchbotDevice):
             return None
 
         _position = max(min(_data[6], 100), 0)
+        _direction_adjusted_position = (100 - _position) if self._reverse else _position
+        _previous_position = self._get_adv_value("position")
+        _in_motion = bool(_data[5] & 0b01000011)
+        self._update_motion_direction(_in_motion, _previous_position, _direction_adjusted_position)
+        
         return {
             "battery": _data[1],
             "firmware": _data[2] / 10.0,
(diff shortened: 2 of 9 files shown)


switchbot/devices/curtain.py
```diff
@@ -54,6 +57,18 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        self._is_closing: bool = False
+
+    def _set_parsed_data(
+        self, advertisement: SwitchBotAdvertisement, data: dict[str, Any]
+    ) -> None:
+        """Set data."""
+        _new_data = replace(
+            advertisement, data=self._sb_adv_data.data | {"data": data}
+        )
+        self._update_motion_direction(_new_data.data["data"].get("inMotion"), self._get_adv_value("position"), _new_data.data["data"].get("position"))
```
bdraco, line 70: Why create new data when we throw it away after the next line?

requirements_dev.txt
```diff
@@ -1,5 +1,6 @@
 pytest-asyncio
 pytest-cov
+pytest-mock
```
bdraco, line 3: I'm not sure this is needed as python is new enough to use `unittest.mock` instead

dcmeglio: Not familiar with that library, will look it up and give it a go.

switchbot/devices/curtain.py
dcmeglio: This was copy/paste from elsewhere. I wasn't exactly sure what this replace was doing if I'm being honest but it seemed like it was merging the advertisement with the existing values from _sb_adv_data. And since I am using data, I thought this might be necessary.  Happy to learn I'm wrong since you know a lot more about python than I do, I'm a C# guy by trade.

bdraco, line 70:
```suggestion
        
        _new_data = replace(
            advertisement, data=self._sb_adv_data.data | {"data": data}
        )
        self._update_motion_direction(_new_data.data["data"].get("inMotion"), self._get_adv_value("position"), _new_data.data["data"].get("position"))
```
```suggestion
        in_motion = data["inMotion"]
        previous_position = self._get_adv_value("position")
        new_position = data["position"]
        self._update_motion_direction(in_motion, previous_position, new_position)
```

dcmeglio: I reread the documentation and I think you're saying I just need something more simple like this?
```
_new_data = self._sb_adv_data.data | {"data": data}
self._update_motion_direction(_new_data["data"].get("inMotion"), self._get_adv_value("position"), _new_data["data"].get("position"))
```
?

dcmeglio: Ok I see what you're saying. My understanding was I need something like `self._sb_adv_data.data | {"data": data}` because reading through the code sometimes we get partial advertisements? Meaning it's possible the position value is missing entirely from the most recent advertisement and the union makes it so we get the existing value from sb_adv_data. E.g. maybe the current advertisement is wholesale missing the inMotion value, so we assume, in its absence, the value is the same from the last advertisement. I'm not finding the comment where I could swear I read this so maybe I hallucinated it. You know the library much better than I do. Let me know if what I said makes sense and if the change I mentioned in https://github.com/Danielhiversen/pySwitchbot/pull/226#discussion_r1446831874 makes sense otherwise I'll accept your suggestion.

bdraco: That seems like it would be a bug in the firmware to send invalid advertisements.  I'm pretty sure it wouldn't get to this point as the parser would reject it if its was short bytes

switchbot/devices/curtain.py
```diff
@@ -121,10 +147,21 @@ async def get_basic_info(self) -> dict[str, Any] | None:
...
+            "inMotion": _in_motion,
+            "position": _direction_adjusted_position,
             "timers": _data[7],
         }
+    
+    def _update_motion_direction(self, in_motion: bool, previous_position: int|None, new_position: int) -> None:
+        """Update opening/closing status based on movement."""
+        if previous_position is None:
+            return
+        if in_motion == True:
```
bdraco, line 159:
```suggestion
        if in_motion is True:
```

Use identity checks for singletons

switchbot/devices/curtain.py
```diff
@@ -121,10 +147,21 @@ async def get_basic_info(self) -> dict[str, Any] | None:
...
+            if new_position != previous_position:
+                self._is_opening = new_position > previous_position
+                self._is_closing = new_position < previous_position
+        else:
+            self._is_closing = self._is_opening = False
```
bdraco, line 164: Reverse the conditional above, set this and return than outdent the above

switchbot/devices/curtain.py
```diff
@@ -210,3 +247,11 @@ def is_calibrated(self) -> Any:
...
         # To get actual light level call update() first.
         return self._get_adv_value("calibration")
+
+    def is_opening(self) -> Any:
+        """Return True if the curtain is opening."""
+        return self._is_opening
+    
+    def is_closing(self) -> Any:
+        """Return True if the curtain is closing."""
+        return self._is_closing
```
bdraco, line 257:
```suggestion
    @property
    def is_opening(self) -> Any:
        """Return True if the curtain is opening."""
        return self._is_opening

    @property
    def is_closing(self) -> Any:
        """Return True if the curtain is closing."""
        return self._is_closing
```

dcmeglio: I'm fine with this change but I was trying to follow the existing pattern from is_reversed and is_calibrated. So this would now be inconsistent with the rest of the pattern.

bdraco: OK

bdraco, line 257:
```suggestion
    def is_opening(self) -> bool:
        """Return True if the curtain is opening."""
        return self._is_opening
    
    def is_closing(self) -> bool:
        """Return True if the curtain is closing."""
        return self._is_closing
```

tests/test_curtain.py
```diff
@@ -0,0 +1,197 @@
...
+    curtain_device._get_basic_info = Mock(side_effect=custom_implementation)
+
+    await curtain_device.get_basic_info()
+
+    basic_info = bytes([0,0,0,0,0,0,90,0])
+
+    await curtain_device.get_basic_info()
+
+    assert curtain_device.is_opening() == False
+    assert curtain_device.is_closing() == False
```
bdraco, line 197: Please use identity checks for `True`/`False` everywhere

dcmeglio:
I think I have addressed all comments.

bdraco:
I'll try to find some time to test this locally make sure it doesn't regress anything obvious

dcmeglio:
Two notes I should add - I only own curtain 3 devices so I didn't test curtain 2. Also I don't have any that open from the center, but I did test both open left and open right with actual devices in addition to my unit tests that hopefully cover that.

Scenarios I tested:
- Sending commands from the Switchbot mobile app and check HA reflected those updates
- Send commands from HA and make sure they're reflected
- Manually moved the curtain to ensure that's reflected
- Opening, stopping mid open
- Closing, stop mid close
- Set position to a device that is "more open" than current state
- Set a position that is "more closed" than current state

During a bunch of that testing I added temp debug logging during any update (advertisement or via active connection) so I could see any changes to the variables to make sure they were never getting set incorrectly.

I didn't spot any issues but more testing is appreciated.

bdraco:
I think I only have curtain v1s

They are in a box somewhere that I need to dig up

bdraco:
Looks like I put them in storage so I'll have to go dig for them

bdraco:
Found it at the very back of storage, of course 🙈

bdraco:
Looks like they are version two, completely dead so I'm charging them

bdraco:
I can't seem to get the curtains added with this PR.  It might be that they need to charge overnight since they have been in storage for so long.  I'll give it a try tomorrow.

In the mean time I fixed the coverage reporting in the CI

bdraco (approved):

bdraco:
@dcmeglio 0.44.0 published.

Please bump the lib HA in a separate PR from the new feature

dcmeglio:
Bump PR submitted https://github.com/home-assistant/core/pull/107833 I'll adjust my existing PR to use the new library once that is merged.

bdraco:
@dcmeglio It looks like this PR caused a regression with blind tilts

https://github.com/home-assistant/core/issues/109947

dcmeglio:
> @dcmeglio It looks like this PR caused a regression with blind tilts
> 
> [home-assistant/core#109947](https://github.com/home-assistant/core/issues/109947)

I'll take a look but I don't own any tilt blinds to test with

bdraco:
Sadly I'm in the same boat on that.   No tilt devices. 

I think we need to make a new base class for curtain ands tilt to inherit from which has the original methods in a cover_device.py

Than curtain and tilt can inherit from that new base class with the original code.  Than have the code overridden only in curtain

dcmeglio:
Agreed. That's along the lines of what I was thinking. And also will need to add unit tests. I added a bunch for the curtain device as part of my PR, but this library is lacking a lot of tests and there are none for tilt blinds really. I should have time to fix it this weekend so hopefully we can include in 2024.2.2. I'll also try to expand some unit test coverage :(

bdraco:
Thanks anything you can do would be great, as this library is mostly on life-support. 

I don't use SwitchBot in production and have only been doing small fixes when I have some spare time.

dcmeglio:
I'm going to order a couple tilts as well (Guess I get to automate my garage blinds now for no real reason???) but their website is currently telling me they can't ship to the US which doesn't sound right

dcmeglio:
Order placed, I'll also see if I can track down https://github.com/home-assistant/core/issues/108087 once I have the devices in hand (sometime next week)

bdraco:
I ordered one for testing as well

Tntdruid:
@bdraco you should try a send a mail to mailto:support@switch-bot.com see if they can send you one for free.

dcmeglio:
@bdraco Still very much a WIP, I have done no testing yet, but wanted to see if this was along the lines you were thinking? https://github.com/Danielhiversen/pySwitchbot/pull/231

PR #230, opened 2024-01-24
Add support for data with zero values from WoSensorTH

sheldonip:
This PR is to support processing data with zero values from WoSensorTH. I have experienced that when the power supply is not stable, the meter would advertise zero values and its screen would be flashing. This PR is to ignore those zero values as if no data is received.

Test case is added for the above case.

### DIFF
--- a/switchbot/adv_parsers/meter.py
+++ b/switchbot/adv_parsers/meter.py
@@ -28,7 +28,7 @@ def process_wosensorth(data: bytes | None, mfr_data: bytes | None) -> dict[str, 
     _temp_f = (_temp_f * 10) / 10
     humidity = temp_data[2] & 0b01111111
 
-    if _temp_c == -0.0 and humidity == 0 and battery == 0:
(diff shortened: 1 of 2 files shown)


switchbot/adv_parsers/meter.py
```diff
@@ -26,13 +26,17 @@ def process_wosensorth(data: bytes | None, mfr_data: bytes | None) -> dict[str,
     )
     _temp_f = (_temp_c * 9 / 5) + 32
     _temp_f = (_temp_f * 10) / 10
+    humidity = temp_data[2] & 0b01111111
+
+    if _temp_c == -0.0 and humidity == 0 and battery == 0:
```
bdraco, line 31:
```suggestion
    if _temp_c == 0 and humidity == 0 and battery == 0:
```

-0 == 0

bdraco (approved):

bdraco:
https://github.com/Danielhiversen/pySwitchbot/releases/tag/0.44.1

PR #231, opened 2024-02-10
Refactored common curtain/tilt code to shared base class

dcmeglio:
This is still a WIP (haven't tested at all yet except running existing unit tests) but wanted to see if this was along the lines you were thinking before I go too far down this road.

### DIFF
--- a/switchbot/devices/base_cover.py
+++ b/switchbot/devices/base_cover.py
@@ -5,6 +5,7 @@ import logging
 from abc import abstractmethod
 from typing import Any
 
+from ..models import SwitchBotAdvertisement
 from .device import REQ_HEADER, SwitchbotDevice, update_after_operation
 
 # Cover keys
@@ -38,6 +39,8 @@ class SwitchbotBaseCover(SwitchbotDevice):
         self._settings: dict[str, Any] = {}
         self.ext_info_sum: dict[str, Any] = {}
         self.ext_info_adv: dict[str, Any] = {}
+        self._is_opening: bool = False
+        self._is_closing: bool = False
 
     async def _send_multiple_commands(self, keys: list[str]) -> bool:
         """Send multiple commands to device.
@@ -115,3 +118,25 @@ class SwitchbotBaseCover(SwitchbotDevice):
             }
 
         return self.ext_info_adv
+
+    def get_light_level(self) -> Any:
+        """Return cached light level."""
+        # To get actual light level call update() first.
+        return self._get_adv_value("lightLevel")
+
+    def is_reversed(self) -> bool:
+        """Return True if curtain position is opposite from SB data."""
+        return self._reverse
+
+    def is_calibrated(self) -> Any:
+        """Return True curtain is calibrated."""
+        # To get actual light level call update() first.
+        return self._get_adv_value("calibration")
+
+    def is_opening(self) -> bool:
+        """Return True if the curtain is opening."""
+        return self._is_opening
+
+    def is_closing(self) -> bool:
+        """Return True if the curtain is closing."""
+        return self._is_closing
--- a/switchbot/devices/base_cover.py
+++ b/switchbot/devices/base_cover.py
@@ -1,9 +1,8 @@
 """Library to handle connection with Switchbot."""
 from __future__ import annotations
 
-from abc import abstractmethod
-
 import logging
+from abc import abstractmethod
 from typing import Any
 
 from .device import REQ_HEADER, SwitchbotDevice, update_after_operation
@@ -67,7 +66,7 @@ class SwitchbotBaseCover(SwitchbotDevice):
                 f"{POSITION_KEYS[1]}{speed:02X}{position:02X}",
             ]
         )
-    
+
     @abstractmethod
     def get_position(self) -> Any:
         """Return current device position."""
--- a/switchbot/devices/base_cover.py
+++ b/switchbot/devices/base_cover.py
@@ -31,10 +31,11 @@ _LOGGER = logging.getLogger(__name__)
 class SwitchbotBaseCover(SwitchbotDevice):
     """Representation of a Switchbot Cover devices for both curtains and tilt blinds."""
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
+    def __init__(self, reverse: bool, *args: Any, **kwargs: Any) -> None:
         """Switchbot Cover device constructor."""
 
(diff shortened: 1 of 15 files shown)


bdraco:
Seems like a reasonable path forward

switchbot/devices/curtain.py
```diff
@@ -168,8 +148,8 @@ def _update_motion_direction(
             self._is_closing = new_position < previous_position
 
     async def get_extended_info_summary(self) -> dict[str, Any] | None:
```
dcmeglio, line 150: I purposely didn't add unit tests for this method because it smells wrong. The code that checks `data[2]` but then only uses `data[1]` just seems wrong, but I don't have any paired curtain devices to test. I don't want to add test cases for code I suspect might be incorrect (but I can't test).

tests/test_blind_tilt.py
```diff
@@ -0,0 +1,126 @@
...
+        (True,bytes([0,1,10,2,0,1,50,4]),[1,1,0,0,1,False,True,False,True,50,4])
+    ]
+)
+async def test_get_basic_info(reverse_mode,data,result):
+    blind_device = create_device_for_command_testing(reverse_mode=reverse_mode)
+
+    async def custom_implementation():
+        return data
+
+    blind_device._get_basic_info = Mock(side_effect=custom_implementation)
```
bdraco, line 88:
```suggestion
    blind_device._get_basic_info = AsyncMock(return_value=data)
```

tests/test_blind_tilt.py
```diff
@@ -0,0 +1,126 @@
...
+        (True,bytes([0,1,10,2,0,0,50,4]),[1,1,0,0,0,False,False,False,False,50,4]),
+        (True,bytes([0,1,10,2,0,1,50,4]),[1,1,0,0,1,False,True,False,True,50,4])
+    ]
+)
+async def test_get_basic_info(reverse_mode,data,result):
+    blind_device = create_device_for_command_testing(reverse_mode=reverse_mode)
+
+    async def custom_implementation():
+        return data
+
```
bdraco, line 87:
```suggestion
```

tests/test_blind_tilt.py
```diff
@@ -0,0 +1,126 @@
...
+    ble_device = generate_ble_device("aa:bb:cc:dd:ee:ff", "any")
+    curtain_device = blind_tilt.SwitchbotBlindTilt(ble_device, reverse_mode=reverse_mode)
+    curtain_device.update_from_advertisement(
+        set_advertisement_data(ble_device, True, position, calibration)
+    )
+    curtain_device._send_multiple_commands = AsyncMock()
+    curtain_device.update = AsyncMock()
+    return curtain_device
+
+def set_advertisement_data(ble_device: BLEDevice, in_motion: bool, position: int, calibration: bool = True):
```
bdraco, line 23:
```suggestion
def make_advertisement_data(ble_device: BLEDevice, in_motion: bool, position: int, calibration: bool = True):
```

maybe `make`?

switchbot/devices/base_cover.py
```diff
@@ -0,0 +1,120 @@
...
+            [
+                f"{POSITION_KEYS[0]}{position:02X}",
+                f"{POSITION_KEYS[1]}{speed:02X}{position:02X}",
+            ]
+        )
+    
+    @abstractmethod
+    def get_position(self) -> Any:
+        """Return current device position."""
+        pass
```
bdraco, line 73:
```suggestion
```

unneeded pass statement

switchbot/devices/base_cover.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @abstractmethod
+    async def get_basic_info(self) -> dict[str, Any] | None:
+        """Get device basic settings."""
+        pass
```
bdraco, line 78:
```suggestion
```

unneeded pass statement

switchbot/devices/base_cover.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @abstractmethod
+    async def get_extended_info_summary(self) -> dict[str, Any] | None:
+        """Get extended info for all devices in chain."""
+        pass
```
bdraco, line 83:
```suggestion
```

unneeded pass statement

switchbot/devices/curtain.py
bdraco: That does look wrong.  I guess we never call this in HA?

dcmeglio: Doesn't look like `get_extended_info_summary` is used at all by HA

dcmeglio:
Other than that assumed bug which I'm now almost positive about looking at `get_extended_info_adv` I've added code coverage and unit tests for my changes. I'm going to work on some formatting cleanups and also still need to test with some actual devices. Otherwise I think this should fix the issue.

bdraco:
You can probably do

```
isort switchbot
black switchbot
ruff --fix switchbot
```

bdraco:
I'll test this tomorrow

switchbot/devices/curtain.py
dcmeglio: https://github.com/OpenWonderLabs/SwitchBotAPI-BLE/blob/latest/devicetypes/curtain.md seems to confirm this is a bug so I'm going to fix that too. Not used by HA, but still a bug in the library.

dcmeglio:
My tilt won't be here for a few days so I can't test that, but my curtains still work fully and the is_opening/is_closing are still fine too. I'm going to consider this ready for review. Let me know how your testing goes. It should just require a version bump in HA.

Technically this is a breaking change in the API since the base class for the Tilt has been changed? If someone was relying on Tilt inheriting from Curtain, this will now cause an issue.

bdraco:
> Technically this is a breaking change in the API since the base class for the Tilt has been changed? If someone was relying on Tilt inheriting from Curtain, this will now cause an issue.

That's a stretch for a breaking change. Should be ok

dcmeglio:
People are reporting it still happens with this code. I’m not spotting the issue right now. Might have to wait till my tilt arrives

bdraco:
I did some adjustments. I think the blind tilt now has all the previous methods so it should work now

bdraco:
Device just showed up

Charging it now

Will test shortly.

bdraco:
Its working with the latest changes I pushed

bdraco:
Thanks @dcmeglio

PR #235, opened 2024-05-22
feat: support users from different regions in obtaining lock encryption key.

huyuwei1996:


huyuwei1996:
Closing the PR because the issue has been resolved with the current release.

PR #236, opened 2024-05-23
Fix lock encryption key retrieval

### LINKED ISSUE #234: API 190 Error
Hello,

When running the get_encryption_key script, I get API 190 error. That's the same error that I get when logging into HAs Bluetooth integration. Any ideas?

`pySwitchbot-0.45.0>get_encryption_key.py MAC EMAIL
Password: 
Unexpected status code returned by SwitchBot API: 190`

dsypniewski:
It seems that SwitchBot has made some changes to their accounts and the previous key retrieval method doesn't work for everybody. This is based on @alexschultze and @hsakoh research, and verified myself with the SwitchBot Android application code.
Since cognito authentication is no longer used I also removed the boto3 dependency.

Fixes #234

Edit: I noticed only after creating this PR that there already is one addressing the exact same issue #235

### DIFF
--- a/switchbot/const.py
+++ b/switchbot/const.py
@@ -10,6 +10,14 @@ DEFAULT_RETRY_TIMEOUT = 1
 DEFAULT_SCAN_TIMEOUT = 5
 
 
+class SwitchbotApiError(RuntimeError):
+    """Raised when API call fails.
+
+    This exception inherits from RuntimeError to avoid breaking existing code
+    but will be changed to Exception in a future release.
+    """
+
+
 class SwitchbotAuthenticationError(RuntimeError):
     """Raised when authentication fails.
 
--- a/switchbot/devices/lock.py
+++ b/switchbot/devices/lock.py
@@ -84,34 +84,42 @@ class SwitchbotLock(SwitchbotDevice):
         return lock_info is not None
 
     @staticmethod
-    async def api_request(subdomain: str, path: str, data: dict = None, headers: dict = None):
-        async with aiohttp.ClientSession(headers=headers, timeout=aiohttp.ClientTimeout(total=10)) as session:
-            url = f"https://{subdomain}.{SWITCHBOT_APP_API_BASE_URL}/{path}"
-            async with session.post(url, json=data) as result:
-                if result.status > 299:
-                    raise SwitchbotApiError(
-                        f"Unexpected status code returned by SwitchBot API: {result.status}"
-                    )
+    async def api_request(
+            session: aiohttp.ClientSession, subdomain: str, path: str, data: dict = None, headers: dict = None
+    ) -> dict:
+        url = f"https://{subdomain}.{SWITCHBOT_APP_API_BASE_URL}/{path}"
+        async with session.post(url, json=data, headers=headers) as result:
+            if result.status > 299:
+                raise SwitchbotApiError(
+                    f"Unexpected status code returned by SwitchBot API: {result.status}"
+                )
 
-                response = await result.json()
-                if response["statusCode"] != 100:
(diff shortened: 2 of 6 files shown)


switchbot/devices/lock.py
```diff
@@ -86,77 +83,78 @@ async def verify_encryption_key(
 
         return lock_info is not None
 
+    @staticmethod
+    def api_request(subdomain: str, path: str, data: dict, headers: dict = None):
```
bdraco, line 87: Since the rest of this library is asyncio, I think this would be a good opportunity to switch to aiohttp instead as there isn't anything that specifically needs requests or a sync library anymore

switchbot/devices/lock.py
```diff
@@ -86,77 +83,78 @@ async def verify_encryption_key(
...
-            auth_response is None
-            or "AuthenticationResult" not in auth_response
-            or "AccessToken" not in auth_response["AuthenticationResult"]
-        ):
-            raise SwitchbotAuthenticationError("Unexpected authentication response")
+        try:
+            userinfo = SwitchbotLock.api_request(
+                "account", "account/api/v1/user/userinfo", {}, auth_headers
+            )
+            region = userinfo["botRegion"]
```
huyuwei1996, line 133: It is recommended that you use "us" when this field is an empty string; otherwise, the domain name will not be able to make requests.

switchbot/devices/lock.py
```diff
@@ -84,34 +84,34 @@ async def verify_encryption_key(
...
-            )
-
-        response = json.loads(result.content)
-        if response["statusCode"] != 100:
-            raise SwitchbotApiError(
-                f"{response['message']}, status code: {response['statusCode']}"
-            )
-
-        return response["body"]
+    async def api_request(subdomain: str, path: str, data: dict = None, headers: dict = None):
```
bdraco, line 87: Please allow passing in an existing session (Home Assistant needs to do this so it doesn't make a new ssl context)

bdraco:
Looks good.  Once last tweak request above.   Its getting close to midnight so I'll test this in the morning when I wake up and get a new release done tomorrow if everything looks good.

bdraco:
Sadly, the batteries are dead in my Test lock so I'll have to go to the store to get new ones

bdraco:
found compatible batteries in another device.

testing now

bdraco (approved):

switchbot/const.py
```diff
@@ -10,6 +10,14 @@
 DEFAULT_SCAN_TIMEOUT = 5
 
 
+class SwitchbotApiError(RuntimeError):
```
bdraco, line 13: The new exception needs to be exported at top level. I'll fix it before I do the release

bdraco:
Thanks @dsypniewski

switchbot/devices/lock.py
```diff
@@ -86,77 +83,97 @@ async def verify_encryption_key(
...
+    @staticmethod
+    async def api_request(
+        session: aiohttp.ClientSession,
+        subdomain: str,
+        path: str,
+        data: dict = None,
+        headers: dict = None,
+    ) -> dict:
+        url = f"https://{subdomain}.{SWITCHBOT_APP_API_BASE_URL}/{path}"
+        async with session.post(url, json=data, headers=headers) as result:
```
bdraco, line 95: Didn't see it till now but I just realized the timeout is removed, and the default timeout here is 300s if not passed. That might be a bit long.

dsypniewski: I agree, I thought since the session is passed from outside it's going to be taken care of there but if that's not the case we should add one here. Here is a PR for that #238

PR #241, opened 2024-06-07
Add basic support for lock pro

szclsya:
Note that since I've only tried manually creating `SwitchbotLock` interface, the `adv_parser` part and the model name is mostly made up for now.

Also since we only need to change some constants to make it work, I just added a enum inside the `SwitchbotLock` to differentiate between the two models.


szclsya:
Looks like the third last number in the lock/unlock command (i.e 0f4e0101000`1`80) encodes the origin of such command:

+ 0: Unknown (Shows up as blank in the app)
+ 1: App Control
+ 2: Widget
+ 3: Apple Watch
+ 4: Scene
+ 5: NFC
+ 6: OpenAPI
+ 7: Unknown
+ 8: Alexa
+ 9: Google Assistant
+ a: SmartThings
+ b: IFTTT
+ c: Matter
+ d: Siri
+ e: Manual
+ f: Keypad

Maybe we should use 6: OpenAPI?

switchbot/devices/lock.py
```diff
@@ -21,10 +22,10 @@
 
 COMMAND_HEADER = "57"
 COMMAND_GET_CK_IV = f"{COMMAND_HEADER}0f2103"
-COMMAND_LOCK_INFO = f"{COMMAND_HEADER}0f4f8101"
-COMMAND_UNLOCK = f"{COMMAND_HEADER}0f4e01011080"
+COMMAND_LOCK_INFO = { "Lock": f"{COMMAND_HEADER}0f4f8101", "LockPro": f"{COMMAND_HEADER}0f4f8102" }
```
bdraco, line 25: Please use an enum for the keys

szclsya: Updated

bdraco:
I removed python 3.10 since it doesn't support StrEnum and fixed the lint issues.

This needs test coverage before merging as well

szclsya:
Added active mode test case. Any idea how can I grab some passive mode data? I ran discover in a loop for a while but only ever got active mode data from the lock.

bdraco:
You can pull it out of diagnostics in Home Assistant from a local Bluetooth adapter or an esphome remote proxy

szclsya:
Hmm weird, from my esphome bluetooth proxy it's still active mode data. Any idea?

schummar:
I have the EU lock and logged my bluetooth traffic. COMMAND_UNLOCK_WITHOUT_UNLATCH should be 0f4e01010001a0

bdraco:
> Hmm weird, from my esphome bluetooth proxy it's still active mode data. Any idea?

What does your yaml look like?

szclsya:
> > Hmm weird, from my esphome bluetooth proxy it's still active mode data. Any idea?
> 
> What does your yaml look like?

Here's diagnostics data from esphome bluetooth proxy and bluetooth integration (identical), which are both in active mode:
```json
{
  "name": null,
  "address": "REDACTED",
  "rssi": -71,
  "advertisement_data": [
    null,
    {
      "2409": {
        "__type": "<class 'bytes'>",
        "repr": "b'\\xc8\\xf5,\\xd9-V\\x1e\\x80\\x00d\\x00\\x00'"
      }
    },
    {
      "0000fd3d-0000-1000-8000-00805f9b34fb": {
        "__type": "<class 'bytes'>",
        "repr": "b'$\\x80d'"
      }
    },
    [],
    -127,
    -71,
    []
  ],
  "details": {
    "source": "REDACTED",
    "address_type": 1
  }
},
```

szclsya:
Okay I configured passive BLE monitor on my ha instance and got this:

```
BLE advertisement received from MAC/UUID REDACTED: service data: []manufacturer specific data: [b'\x0f\xffi\t\xc8\xf5,\xd9-V6\x82\x00d\x00\x00']local name: UUID16: None,UUID128: None
```

It's not currently parsable by the adv_parser though. At line `if mfr_id in advertisement_data.manufacturer_data:` there's `ValueError: byte must be in range(0, 256)`.

szclsya:
Should be good now.

switchbot/devices/lock.py
```diff
@@ -46,6 +60,7 @@ class SwitchbotLock(SwitchbotDevice):
     def __init__(
         self,
         device: BLEDevice,
+        model: LockModel,
```
bdraco, line 63: This is a breaking change. It could be a kwarg instead

switchbot/devices/lock.py
```diff
@@ -64,15 +79,16 @@ def __init__(
         self._key_id = key_id
         self._encryption_key = bytearray.fromhex(encryption_key)
         self._notifications_enabled: bool = False
+        self._model: LockModel = model
         super().__init__(device, None, interface, **kwargs)
 
     @staticmethod
     async def verify_encryption_key(
-        device: BLEDevice, key_id: str, encryption_key: str
+        device: BLEDevice, model: LockModel, key_id: str, encryption_key: str
```
bdraco, line 87: This is a breaking change. It could be a kwarg instead

switchbot/devices/lock.py
szclsya: Moved model to kwargs, default to the original lock.

switchbot/devices/lock.py
```diff
@@ -64,15 +78,20 @@ def __init__(
...
+        self._model: LockModel = kwargs.get("model") or LockModel.LOCK
         super().__init__(device, None, interface, **kwargs)
 
     @staticmethod
     async def verify_encryption_key(
-        device: BLEDevice, key_id: str, encryption_key: str
+        device: BLEDevice,
+        key_id: str,
+        encryption_key: str,
+        **kwargs: Any,
```
bdraco, line 89: model can go before the **kwargs unpack

szclsya: Updated

bdraco:
I think this is close to ready to merge.

Do you have a Home Assistant PR to match up to this?

I'll try to take a look tomorrow, but it might not be until Tuesday that I come back to this as I have a busy day tomorrow and its super late here.

ynsgnr:
Thank you so much for the work @szclsya, tested it with EU Lock Pro and it works well! Unlocking, locking, status and night latch unlocking too!

I think model detection can be improved by detecting Lock or Lock Pro automatically. Probably by using the same enum (`SwitchbotModel`) instead of two different enums (`SwitchbotModel` , `LockModel`)

That would allow a lock construction like this:

`lock.SwitchbotLock(wolock_data.device, "....", "....", model=wolock_data.data["modelName"])`

incre77:
Thank you very much for your work. Im have been testing but in my case is not fully working.
I made a guest user and i got the encryption key for that user (cos i began with the main account in google third party mode and its imposible to me to reset password or get the encryption key for that account).

I tried this:
```
async def main():
    wolock = await GetSwitchbotDevices().get_locks()
    print(wolock)
    l = lock.SwitchbotLock(wolock['FF:02:53:A3:BB:11'].device,"13", "XXXXXXXXXXXXXXXXX","$")
    await l.get_basic_info()
    await l.unlock()
    
asyncio.run(main())
```
And the lock only "locks".
If the lock is unlocked the lock locks whenever i send lock or unlock command but if the lock is locked the lock make a "beep" but nothing happens.
the result of get_basic_info is alwais = bytearray(b'\x05') = "Unsuccessful, please try again"
and the result of self._send_command(command) when lock or unlock is = bytearray(b'\x01\x82L\x00m\x00') or bytearray(b'\x01\x82Z\x00m\x00') when do nothing and bytearray(b'\x01\x90Z\x00m\x00') when locks.

I supose my problem is using a guest account but maybe its posible to make it work too. (meanwhile im talking with the support to get the main account password)
i have an european lock pro with the latest firmware.

szclsya:
@ynsgnr Good idea! Updated.

szclsya:
@bdraco I have a Home Assistant patch working locally now. Will send a pull request after this pull request is merged.

bdraco:
> @bdraco I have a Home Assistant patch working locally now. Will send a pull request after this pull request is merged.

It would be super helpful for review purposes if you linked the branch, or opened a draft PR to HA core so I can look at the whole change.

switchbot/__init__.py
```diff
@@ -40,6 +40,7 @@
     "SwitchbotApiError",
     "SwitchbotAuthenticationError",
     "ColorMode",
+    "LockModel",
```
bdraco, line 43:
```suggestion
```

szclsya:
@bdraco See Home Assistant PR at home-assistant/core#119326

ynsgnr:
@bdraco gentle reminder to take a look at the PR, just wanted to remind  😇

bdraco:
Haven't forgotten. Just haven't had any free time to look

switchbot/devices/lock.py
```diff
@@ -59,20 +73,27 @@ def __init__(
             raise ValueError("encryption_key is missing")
         elif len(encryption_key) != 32:
             raise ValueError("encryption_key is invalid")
+        if model != SwitchbotModel.LOCK and model != SwitchbotModel.LOCK_PRO:
```
bdraco, line 76:
```suggestion
        if model not in (SwitchbotModel.LOCK, SwitchbotModel.LOCK_PRO):
```

bdraco (approved):
Thanks @szclsya

bdraco:
released as 0.48.0

Please bump the library in a separate PR from https://github.com/home-assistant/core/pull/119326 and than merge in dev to https://github.com/home-assistant/core/pull/119326 once the bump PR is merged

szclsya:
> Please be sure to name future branches something besides `master` as it means I have to do a lot more work to pull them down to review which makes reviews take longer.
> 
> Thanks

Roger that. Sorry for the inconvenience.

PR #245, opened 2024-07-23
possible night latch fix - lock pro

Michal4K:
After analyzing publicly posted lock pro adv data. I think I found the right bit indicating night latch.
But this needs to be tested by someone who actually have lock pro.

update pySwitchbot library in HomeAssistent from my fork
`pip install --upgrade --no-deps --force-reinstall git+https://github.com/Michal4K/pySwitchbot@night_latch_lock_pro`

**Do not forget --no-deps**, unless you want broke up your HA dependencies.
Then restart HA and reload switchbot integration.

So anyone feel free to test this and report back.

PR #247, opened 2024-08-20
disable detection of night latch for lock pro

Michal4K:
We should just disable this. It looks like night latch bit is not anymore in Lock Pro ADV.
Discussed in this draft: https://github.com/Danielhiversen/pySwitchbot/pull/245

For Home Assistant i create a little work around to force lock to be setup as if night latch is detected.
https://github.com/home-assistant/core/pull/124326#issue-2476424411

I don't believe OpenWonderLabs will publish documentation, which would help with correct implementation anytime soon.


bdraco (changes requested):
Please update the adv parser tests

Michal4K:
I added one test based on data from this PR discussion: https://github.com/Danielhiversen/pySwitchbot/pull/243#issuecomment-2219101334

In this case adv parser nightlatch would be enabled, but he didn't have this feature turned on.

bdraco (approved):

PR #252, opened 2024-10-14
Add initial support for the keypad touch

### LINKED ISSUE #250: Feature request - Support Keypad
Is there any chance for support of the keypad? Would be great to disable the keypad or get alerts into HA when the keypad is removed.

nickconway:
Initial support for the keypad touch. I currently have this working with home assistant on my own fork, and I can tell when a successful attempt was done. I'm sure there's more information to pull from it once there's official documentation.

Closes #250


bdraco:
I’d probably add a note in the Home Assistant documentation that there is no encryption, so you shouldn’t rely on the sensor for secure operation

switchbot/adv_parsers/keypad.py
```diff
@@ -0,0 +1,24 @@
...
+from __future__ import annotations
+
+import logging
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def process_wokeypad(
+    data: bytes | None, mfr_data: bytes | None
+) -> dict[str, bool | int]:
```
bdraco, line 12: Typing doesn’t match all the return types

switchbot/adv_parsers/keypad.py
```diff
@@ -0,0 +1,24 @@
+"""Keypad parser."""
+
+from __future__ import annotations
+
+import logging
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def process_wokeypad(
```
bdraco, line 10: Tests are missing for this

switchbot/adv_parsers/keypad.py
```diff
@@ -0,0 +1,24 @@
...
+    data: bytes | None, mfr_data: bytes | None
+) -> dict[str, bool | int | None] | {}:
+    """Process woKeypad services data."""
+    if data is None and mfr_data is None:
+        return {}
+
+    if data is None:
+        return {"battery": None, "attemptState": None}
+
+    _LOGGER.debug("mfr_data: %s", mfr_data.hex())
```
bdraco, line 20: `mfr_data` can be `None` here which would raise when trying to call `.hex()` on it

switchbot/adv_parsers/keypad.py
```diff
@@ -0,0 +1,30 @@
...
+) -> dict[str, bool | int | None]:
+    """Process woKeypad services data."""
+    if data is None or mfr_data is None:
+        return {"battery": None, "attemptState": None}
+
+    _LOGGER.debug("mfr_data: %s", mfr_data.hex())
+    if data:
+        _LOGGER.debug("data: %s", data.hex())
+
+    global lastStatus
```
bdraco, line 23: Please don't use global variables in new code.

switchbot/adv_parsers/keypad.py
```diff
@@ -0,0 +1,32 @@
...
+    """Process woKeypad services data."""
+    if data is None or mfr_data is None:
+        process_wokeypad.lastStatus = -1
+        return {"battery": None, "attemptState": None}
+
+    _LOGGER.debug("mfr_data: %s", mfr_data.hex())
+    if data:
+        _LOGGER.debug("data: %s", data.hex())
+
+    lastStatus = process_wokeypad.lastStatus
```
bdraco, line 22: This won't work if they have multiple keypads. Its effectively the same as a global

nickconway: Yeah I just realized that whoops, I'll move it into a device class

switchbot/adv_parsers/keypad.py
```diff
@@ -0,0 +1,28 @@
...
+
+from __future__ import annotations
+
+import logging
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def process_wokeypad(
+    data: bytes | None, mfr_data: bytes | None, lastStatus: list[int] = [-1]
```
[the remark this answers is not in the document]
nickconway: Ah that makes sense, where would you suggest managing it at? I'm having trouble seeing the data flow after the advertisement parser is called, I've tried overriding methods from the base class but they asen't having any effect.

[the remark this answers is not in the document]
bdraco: Wrap the value as an IntEnum and let Home Assistant track it downstream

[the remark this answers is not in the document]
nickconway: IntEnum or just an int? The value continuously increases, I don't think an enum would work.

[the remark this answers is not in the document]
bdraco: Just an int then

switchbot/adv_parsers/keypad.py
```diff
@@ -0,0 +1,21 @@
...
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def process_wokeypad(
+    data: bytes | None, mfr_data: bytes | None,
+) -> dict[str, bool | int | None]:
+    """Process woKeypad services data."""
+    if data is None or mfr_data is None:
+        return {"battery": None, "attemptState": None}
```
bdraco, line 15:
```suggestion
        return {"battery": None, "attempt_state": None}
```

switchbot/adv_parsers/keypad.py
```diff
@@ -0,0 +1,21 @@
...
+
+    _LOGGER.debug("mfr_data: %s", mfr_data.hex())
+    if data:
+        _LOGGER.debug("data: %s", data.hex())
+
+    return {"battery": data[2], "attemptState": mfr_data[6]}
```
bdraco, line 21:
```suggestion
    return {"battery": data[2], "attempt_state": mfr_data[6]}
```

Please use snake case for new keys

bdraco, line 21: Are you sure battery uses all the bits?
Most of the other devices need  `& 0b01111111`

nickconway: Nothing from the advertisement data besides the attempt state has changed as I've been using it so I don't know if the eigth bit is used for anything but it only needs 7 bits to make 100 anyway, I'll fix this

switchbot/adv_parser.py
```diff
@@ -187,7 +194,6 @@ class SwitchbotSupportedType(TypedDict):
         "modelName": SwitchbotModel.RelaySwitch1Plus,
         "modelFriendlyName": "Relay Switch 1",
         "func": process_worelay_switch_1plus,
```
bdraco, line 196:
```suggestion
        "func": process_worelay_switch_1plus,
        "manufacturer_id": 2409,
```

Issue #246, opened 2024-08-17
feature request: support for the Water Leak Detector


MikeDembek:
Hey, I've gotten a couple of the new Water Leak Detectors and would be interested in support for those.  It looks like there are a number of others in the HA community interested as well:

https://community.home-assistant.io/t/switchbot-bluetooth-water-leak-detector/749332

Here's the API documentation:

https://github.com/OpenWonderLabs/SwitchBotAPI?tab=readme-ov-file#water-leak-detector

Thanks for any help, and if there's more info I can provide I'll try to help out.

luc-ass:
Hey there. Great to see that there is more interest in this. I would love to help out, but can not figure out on where to start. @Danielhiversen could you point me in the right direction? I have a spare sensor laying around...

luc-ass:
@MikeDembek looks like this can be closed, as it has been merged

PR #262, opened 2024-10-30
add Leak Detector

luc-ass:
Hi @Danielhiversen,

I have added the new [Leak Detector](https://eu.switch-bot.com/products/switchbot-water-leak-detector) to pySwitchbot, as discussed in #246. From what I can see this should work but I am not quite sure on how to robustly test, whether all works as expected. It would be great, if you could point me in the right direction.

Thanks,
Lucas


switchbot/adv_parsers/leak.py
```diff
@@ -0,0 +1,33 @@
+def process_leak(data: bytes | None, mfr_data: bytes | None) -> dict[str, bool | int]:
+    """Process SwitchBot Water Leak Detector advertisement data."""
+    if data is None and mfr_data is None:
+        return {}
+
+    water_leak_detected = None
```
bdraco, line 6: Please add tests for this parser

bdraco:
If you run 
`pre-commit install` 
`pre-commit run --all-files`

It should fix the lint issues for you

bdraco (changes requested):
Please add tests for this parser

tests/test_adv_parser.py
```diff
@@ -1737,3 +1737,33 @@ def test_parse_advertisement_data_relay_switch_1():
...
+
+
+def test_parse_advertisement_data_leak():
+    """Test parse_advertisement_data for the leak."""
+    ble_device = generate_ble_device("aa:bb:cc:dd:ee:ff", "any")
+    adv_data = generate_advertisement_data(
+        manufacturer_data={2409: b"\xc4407Lz\x18N\x98g^\x94Q<\x05\x00\x00\x00\x00"},
+        service_data={"0000fd3d-0000-1000-8000-00805f9b34fb": b"&\x00N"},
+        rssi=-72,
+    )
```
bdraco, line 1749: Please add a test for passive scans as well.  In passive mode, the service_data will be empty

bdraco:
> Hi @bdraco thank you for your input! This addition still needs to be tested with my hardware. Have you got a process how to check which response corresponds to which state? I am not super expericened with hardware testing ☺️

Try to get the device to end up in all possible states and write tests to each state to assert that the data reflects the state you believe the device to be in.  When there aren't docs its usually hours of trial and error to get it as close to correct as possible.

luc-ass:
> Try to get the device to end up in all possible states and write tests to each state to assert that the data reflects the state you believe the device to be in. When there aren't docs its usually hours of trial and error to get it as close to correct as possible.

I have now acquired data for leak/no leak and different battery levels. I have no idea how to "tamper" with the device, but that kind of message is supported [here](https://github.com/OpenWonderLabs/node-switchbot/blob/latest/src/device/woleak.ts), so I implemented. For now I was able to test with a battery at 9% but still no low_battery bit, while the app already shows it. Looks like it is not really used.

switchbot/adv_parsers/leak.py
```diff
@@ -0,0 +1,33 @@
...
+    device_tampered = None
+    battery_level = None
+    low_battery = None
+
+    # Check model ID (first byte should be 0x26 for Water Leak Detector)
+    if data[0] != 0x26:
+        return {}
```
bdraco, line 16: Please add coverage for this branch

switchbot/devices/leak.py
```diff
@@ -0,0 +1 @@
+from __future__ import annotations
```
bdraco, line 1: We can drop this file since there is no active connection

bdraco (approved):

bdraco:
Thanks @luc-ass

PR #263, opened 2024-11-15
Add support Relay Switch 1 and Relay Switch 1PM

greyeee:
This pull request is from the in-house dev team at SwitchBot. Two new models of relay switches are added, SwitchBot Relay Switch 1 and Relay Switch 1PM.


bdraco (approved):

bdraco:
Missed one thing in review

fixed in https://github.com/Danielhiversen/pySwitchbot/pull/266

PR #272, opened 2024-11-25
Fix the voltage value of the relay switch

greyeee:
Fix the voltage value of the relay switch, combine the advertisement and obtain the voltage and current through the command to update the device state.

### DIFF
--- a/switchbot/devices/relay_switch.py
+++ b/switchbot/devices/relay_switch.py
@@ -1,3 +1,5 @@
+import asyncio
+import logging
 import time
 from typing import Any
 
@@ -5,7 +7,10 @@ from bleak.backends.device import BLEDevice
 from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
 
 from ..const import SwitchbotModel
-from .device import SwitchbotSequenceDevice
+from ..models import SwitchBotAdvertisement
+from .device import SwitchbotDevice
+
+_LOGGER = logging.getLogger(__name__)
 
 COMMAND_HEADER = "57"
 COMMAND_GET_CK_IV = f"{COMMAND_HEADER}0f2103"
@@ -16,7 +21,7 @@ COMMAND_GET_VOLTAGE_AND_CURRENT = f"{COMMAND_HEADER}0f7106000000"
 PASSIVE_POLL_INTERVAL = 1 * 60
 
 
-class SwitchbotRelaySwitch(SwitchbotSequenceDevice):
+class SwitchbotRelaySwitch(SwitchbotDevice):
     """Representation of a Switchbot relay switch 1pm."""
 
     def __init__(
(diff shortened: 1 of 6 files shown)


switchbot/devices/relay_switch.py
```diff
@@ -43,6 +48,28 @@ def __init__(
...
+        new_state = self._get_adv_value("sequence_number")
+        _LOGGER.debug(
+            "%s: update advertisement: %s (seq before: %s) (seq after: %s)",
+            self.name,
+            advertisement,
+            current_state,
+            new_state,
+        )
+        if current_state != new_state:
+            asyncio.ensure_future(self.update())
```
bdraco, line 71: We should use create_task here instead

Additionally the event loop doesn't hold strong references to tasks so we need to hold one ourselves to ensure it's not garbage collected before it finishes

We also need to handle the case where a previous update is still in progress and the task isn't finished yet

greyeee: Thank you for your valuable advice. I have addressed the issues you pointed out. I have committed the changes and pushed them to the repository. Could you please review the updated code at your earliest convenience?

bdraco: We still have a risk of multiple updates running at once

A better solution is to force `poll_needed` to be `True` and let Home Assistant's coordinator manage the poll in https://github.com/home-assistant/core/blob/5ef5838b20d4d6f8d340a008cb7004bb4b580228/homeassistant/components/bluetooth/active_update_coordinator.py#L168

```diff
diff --git a/switchbot/devices/relay_switch.py b/switchbot/devices/relay_switch.py
index 3c7272b..681f270 100644
--- a/switchbot/devices/relay_switch.py
+++ b/switchbot/devices/relay_switch.py
@@ -46,6 +46,7 @@ class SwitchbotRelaySwitch(SwitchbotDevice):
         self._key_id = key_id
         self._encryption_key = bytearray.fromhex(encryption_key)
         self._model: SwitchbotModel = model
+        self._force_next_update = False
         super().__init__(device, None, interface, **kwargs)
 
     def update_from_advertisement(self, advertisement: SwitchBotAdvertisement) -> None:
@@ -68,7 +69,7 @@ class SwitchbotRelaySwitch(SwitchbotDevice):
             new_state,
         )
         if current_state != new_state:
-            asyncio.create_task(self.update())
+            self._force_next_update = True
 
     async def update(self, interface: int | None = None) -> None:
         """Update state of device."""
@@ -90,6 +91,9 @@ class SwitchbotRelaySwitch(SwitchbotDevice):
 
     def poll_needed(self, seconds_since_last_poll: float | None) -> bool:
         """Return if device needs polling."""
+        if self._force_next_update:
+            self._force_next_update = False
+            return True
         if (
             seconds_since_last_poll is not None
             and seconds_since_last_poll < PASSIVE_POLL_INTERVAL

```

greyeee: This is a very good suggestion, I have already made the changes according to your suggestion. Could you please review the updated code at your earliest convenience?

PR #281, opened 2024-12-22
Fix Leak Detector model `3` → `&`, add tests from HA

luc-ass:
This pull request:
- fixes the model/type of the Leak Detector from `3` to `&`
- fixes the old tests
- adds a test with real world data from Home Assistant
- adds an `async def get_leakdetectors()`

It has now been tested with Home Assistant. A corresponding Pull Request will be opened soon.


luc-ass:
Looks to me like the coverage problem of `discovery.py` is not real, as there is no dedicated `test_discovery.py`. Isn't that implicit?

bdraco:
> Looks to me like the coverage problem of `discovery.py` is not real, as there is no dedicated `test_discovery.py`. Isn't that implicit?

The coverage is truly missing. There isn't any coverage for this file

bdraco (approved):

PR #294, opened 2025-01-30
Add support for WoRemote

awahlig:
This adds support for [Switchbot Remote](https://switch-bot.com/products/switchbot-remote) (the small, 2-button remote).

Can (only) be used to monitor the battery level.

I'd like to use this in Home Assistant to notify me when the battery gets low (I have HA support implemented as well but this would need to go in first).

The device isn't documented in the [Switchbot BLE API](https://github.com/OpenWonderLabs/SwitchBotAPI-BLE/tree/latest) but the battery level is in the first few bytes of advertisement data which is the same for all device types.

### DIFF
--- a/switchbot/adv_parser.py
+++ b/switchbot/adv_parser.py
@@ -29,6 +29,7 @@ from .adv_parsers.relay_switch import (
     process_worelay_switch_1,
     process_worelay_switch_1pm,
 )
+from .adv_parsers.remote import process_woremote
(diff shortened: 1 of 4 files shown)


bdraco (changes requested):
Please add tests for the advertisement parser for both active and passive scans

bdraco (approved):

bdraco:
Thanks @awahlig

PR #296, opened 2025-02-02
Add support for Evaporative Humidifier

dsypniewski:
This adds support for Evaporative Humidifier (without docking station) in preparation to start working on HA integration.


bdraco:
> Or maybe should I do this in this PR as well?

It would be best to do a PR to extract them first, than we can back merge it into this PR which would make it a lot easier to review since it would minimize the amount of code changes

bdraco:
I noticed a few things that the linter wasn't picking up when doing the review so I turned on some more rules that would catch these automatically as its likely easier than having to wait for me to review.

I think this is ready to go once the issues ruff has found are addressed.

tests/test_evaporative_humidifier.py
```diff
@@ -0,0 +1,206 @@
...
+        await device.set_mode(0)
+        assert False
+    except ValueError:
+        assert True
+
+    try:
+        await device.set_mode(HumidifierMode.TARGET_HUMIDITY)
+        assert False
+    except TypeError:
+        assert True
```
bdraco, line 154: Use `pytest.raises` with a match for these instead

switchbot/const.py
```diff
@@ -71,3 +72,35 @@ class LockStatus(Enum):
...
+    HumidifierMode.LOW,
+    HumidifierMode.MEDIUM,
+    HumidifierMode.HIGH,
+}
+
+
+TARGET_HUMIDITY_MODES = {
+    HumidifierMode.SLEEP,
+    HumidifierMode.TARGET_HUMIDITY,
+}
```
bdraco, line 106: I'd keep these in the humidifier module since they aren't used outside of it and nothing else needs them AFAICT

switchbot/const.py
```diff
@@ -71,3 +72,35 @@ class LockStatus(Enum):
     LOCKING_STOP = 4  # LOCKING_BLOCKED
     UNLOCKING_STOP = 5  # UNLOCKING_BLOCKED
     NOT_FULLY_LOCKED = 6  # LATCH_LOCKED - Only EU lock type
```
bdraco, line 74: not for this PR, but its a bit strange that we have lock data in `const.py` and not in the lock platforms

bdraco (approved):

switchbot/const.py
dsypniewski: Moving them there causes circular dependency issues, how about separate `const_*.py` or `const/*.py` files for different devices?

bdraco: Works for me

dsypniewski: Sorry, I'm not sure if you meant it as moving them to `switchbot/devices/evaporative_humidifier.py` works for you or you agree with with separate const files?

bdraco: Separate constant or models files is fine

switchbot/const.py
dsypniewski: Please take a look at #303 I did extract that part in that PR in order to do the same here

dsypniewski:
I think this should be good to go

switchbot/devices/evaporative_humidifier.py
```diff
@@ -0,0 +1,193 @@
...
+    @classmethod
+    async def verify_encryption_key(
+        cls,
+        device: BLEDevice,
+        key_id: str,
+        encryption_key: str,
+        model: SwitchbotModel = SwitchbotModel.EVAPORATIVE_HUMIDIFIER,
+        **kwargs: Any,
+    ) -> bool:
+        return await super().verify_encryption_key(
```
bdraco, line 62: I'm not sure this will work without implementing `get_basic_info`

https://github.com/sblibs/pySwitchbot/blob/94882d3f611e3d015f98385a36e21e2a0ea1daf8/switchbot/devices/device.py#L813

dsypniewski: Hmm, you are right. The problem is that I don't know what would that command be for this device and what kind of data it should contain. This device doesn't have battery so looking to match that value is out the window. I'll try to come up with something.

bdraco: I'm going to be traveling again soon so I may miss GitHub notifications.  Feel free to reach out on Discord (same handle) if you don't hear back for a bit

bdraco: I think any command that works will probably be fine for now with a comment that we can replace it with something better later

dsypniewski: The latest push should take care of this

bdraco:
Thanks @dsypniewski

PR #326, opened 2025-05-07
Add  support for vacuum

zerzhang:
This adds support for vacuum in preparation to start working on HA integration.


bdraco (approved):
Thanks!

PR #337, opened 2025-05-19
Add support for hub3

zerzhang:
This add support for Hub3 in preparation to start working on HA integration.

Note:  For some new devices, we use the last four bytes in SCAN RSP to represent the device type.


bdraco:
Thanks @zerzhang

PR #340, opened 2025-05-20
Add lock series products

zerzhang:
1. Add lock ultra and lock lite
2. correct some adv parsers data


switchbot/devices/lock.py
```diff
@@ -56,7 +64,12 @@ def __init__(
         model: SwitchbotModel = SwitchbotModel.LOCK,
         **kwargs: Any,
     ) -> None:
-        if model not in (SwitchbotModel.LOCK, SwitchbotModel.LOCK_PRO):
+        if model not in (
```
bdraco, line 67: Would you please add coverage for this?

bdraco (approved):
Thanks @zerzhang

PR #354, opened 2025-06-09
Fix issue the iv is None

zerzhang:
Fix the issues mentioned in https://github.com/home-assistant/core/issues/142316#issuecomment-2954519931

This happens when you perform an action while the device is disconnecting.


switchbot/devices/device.py
```diff
@@ -851,9 +851,6 @@ async def _send_command(
         return result[:1] + self._decrypt(result[4:])
 
     async def _ensure_encryption_initialized(self) -> bool:
-        if self._iv is not None:
-            return True
-
```
bdraco, line 848: I think this needs a more robust fix. We probably need better synchronization to handle cases where a command is executed during or just before disconnection.

PR #362, opened 2025-06-25
Return the snake_case style for led effects

zerzhang:
For the led like bulb/light strip, we return a snake_case effect,  so we can use translations to translate them and we can add icons to them in HA


bdraco (approved):

zerzhang:
> @zerzhang I did some cleanups. Let me know if that works for you

Yeah, I have tried and it works well

PR #368, opened 2025-07-03
Fix race condition when decrypting notifications during lock disconnection

### LINKED ISSUE #353: Lock status abnormal in HA
@bdraco Hi, Could you help to have a look for case https://github.com/home-assistant/core/issues/142316 and https://github.com/home-assistant/core/issues/146364 and give me some suggestion,  The lock works well but but the status is abnormal until the user reload the device

bdraco:
## Fix race condition when decrypting notifications during lock disconnection

Fixes #353 

### Problem
A race condition occurs when encrypted notifications arrive during device disconnection:
1. Encrypted command sent to lock
2. Device starts disconnecting (clears IV)
3. Notification arrives with encrypted response
4. `_decrypt` fails with `RuntimeError: Cannot decrypt: IV is None`

This timing-dependent issue is hard to reproduce and may vary across Bluetooth stacks.

### Solution
Modified `_decrypt` to gracefully handle missing IV during expected disconnection:
- Check `_expected_disconnect` flag before raising error
- Log debug message and return empty bytes during disconnection
- Still raise `RuntimeError` for genuine IV issues

### Testing
Added test `test_decrypt_with_none_iv_during_disconnect` to verify the fix works correctly.

### DIFF
--- a/switchbot/devices/device.py
+++ b/switchbot/devices/device.py
@@ -955,6 +955,12 @@ class SwitchbotEncryptedDevice(SwitchbotDevice):
         if len(data) == 0:
             return b""
         if self._iv is None:
+            if self._expected_disconnect:
+                _LOGGER.debug(
+                    "%s: Cannot decrypt, IV is None during expected disconnect",
+                    self.name,
+                )
+                return b""
             raise RuntimeError("Cannot decrypt: IV is None")
         decryptor = self._get_cipher().decryptor()
         return decryptor.update(data) + decryptor.finalize()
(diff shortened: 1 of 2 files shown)


tests/test_lock.py
```diff
@@ -478,6 +479,34 @@ def test_notification_handler_not_enabled(model: str):
...
+@pytest.mark.parametrize(
+    "model",
+    [
+        SwitchbotModel.LOCK,
+        SwitchbotModel.LOCK_LITE,
+        SwitchbotModel.LOCK_PRO,
+        SwitchbotModel.LOCK_ULTRA,
+    ],
+)
+def test_notification_handler_during_disconnect(
```
bdraco, line 491: We could really use a good end to end test framework, but since we don't have one, this will have to do

PR #372, opened 2025-07-09
Fix periodic reporting status for lock series

zerzhang:
- suggest remove the lock periodic reporting status when perfoming an action lock or unlock, which contains error status and sometimes causes the status to be overwritten immediately
 - Normally, when we perform a locking or unlocking action, we will actively query the lock status once. At this time, the locking or unlocking action may not have been completed. Then, when the BLE scan detects a change in the device broadcast, we will also actively update the status once. Through these two methods, we can ensure that the final state of the lock is correct.

### DIFF
--- a/switchbot/devices/lock.py
+++ b/switchbot/devices/lock.py
@@ -37,7 +37,12 @@ COMMAND_LOCK = {
     SwitchbotModel.LOCK_PRO: f"{COMMAND_HEADER}0f4e0101000000",
     SwitchbotModel.LOCK_ULTRA: f"{COMMAND_HEADER}0f4e0101000000",
 }
-COMMAND_ENABLE_NOTIFICATIONS = f"{COMMAND_HEADER}0e01001e00008101"
+COMMAND_ENABLE_NOTIFICATIONS = {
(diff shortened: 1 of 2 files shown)


zerzhang:
If the user only has passive scanning, the user should not be able to add this device.

bdraco:
Historically users turn on active scanning to add the device and than turn it back off once it's discovered.  When active scans are on the connection times are much much longer

zerzhang:
@bdraco I have discussed this with our internal team and was told that different devices have different instructions for periodically reporting device status, so I made the following changes.
- Correct COMMAND_ENABLE_NOTIFICATIONS command for differenct lock device
- Each time an action is executed, the device status is reported periodically until Bluetooth is disconnected

zerzhang:
@bdraco Could you help to review this?

bdraco:
Thanks for your patience! I'm currently awaiting battery delivery for proper testing of the lock functionality. Due to shipping restrictions on lithium-ion batteries to my location, delivery has been delayed, but tracking indicates arrival within 1-2 days.

Once received, I'll thoroughly test the periodic reporting changes with actual hardware and merge if everything works as expected.

I'll update this PR as soon as testing is complete.

bdraco (approved):

PR #409, opened 2025-11-05
fix climate panel adv parse

zerzhang:
Fixed the parsing of the light property


bdraco:
Should I merge this one and do a patch release before reviewing the other ones and do a feature release after?

bdraco (approved):
Thanks @zerzhang

PR #410, opened 2025-11-06
Add support for presence sensor

zerzhang:
- Add presence sensor

### DIFF
--- a/switchbot/adv_parsers/presence_sensor.py
+++ b/switchbot/adv_parsers/presence_sensor.py
@@ -1,5 +1,7 @@
 """Advertisement data parser for presence sensor devices."""
 
+from __future__ import annotations
+
 import logging
 
 from ..const.presence_sensor import BATTERY_LEVEL_MAP


borgqueenx:
If this is complete, does it mean the presence sensor is supported by switchbot bluetooth integration in home assistant in the next release? @zerzhang

zerzhang:
> @zerzhang Would you please resolve conflicts

fixed

switchbot/adv_parsers/presence_sensor.py
```diff
@@ -0,0 +1,43 @@
+"""Advertisement data parser for presence sensor devices."""
+
```
bdraco, line 2:
```suggestion

from __future__ import annotations

```

lukasmalkmus:
I did some attempt of creating a PR in HomeAssistant but I've never contributed to it, so far. So there might be developers more suited for that :)

But I wonder if in this PR the presence sensor exposes `led_state` when it should probably expose `is_light`?

PR #421, opened 2025-11-29
Fix ENCRYPTION_KEY var in README

LennyPenny:
Fixes var name being unabbreviated when defined.


bluetoothbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - Renaming only the definition reintroduces the NameError the examples already had
> - Same definition/call-site mismatch in the Lock example

PR #454, opened 2026-03-02
Add support for Candle Warmer Lamp

XiaoLing-git:
**1. Candle Warmer Lamp**


switchbot/const/light.py
```diff
@@ -6,6 +6,7 @@ class ColorMode(Enum):
     COLOR_TEMP = 1
     RGB = 2
     EFFECT = 3
+    BRIGHTNESS = 11
```
bdraco: why 11 ?

switchbot/devices/light_strip.py
```diff
@@ -304,6 +304,58 @@ def color_modes(self) -> set[ColorMode]:
...
+    async def get_basic_info(self) -> dict[str, Any] | None:
+        """Get device basic settings."""
+        if not (
+            res := await self._get_multi_commands_results(self._get_basic_info_command)
+        ):
+            return None
+        _version_info, _data = res
+        return {
+            "isOn": bool(_data[1] & 0b10000000),
+            "brightness": _data[2] & 0b01111111,
```
bdraco, line 310: is firmware missing on purpose?

bdraco (changes requested):
Please see comments above

switchbot/const/light.py
XiaoLing-git: I observed other enumeration definitions in the same file, which appear to be related. To avoid potential issues, I assigned a non-existent value to them.

switchbot/devices/light_strip.py
XiaoLing-git: fixed

zerzhang (dismissed):

tests/test_strip_light.py
```diff
@@ -114,13 +128,33 @@ async def test_default_info(device_case, expected_effects):
         assert effect in effect_list
 
 
+@pytest.mark.asyncio
+async def test_candle_warmer_lamp_info():
```
bdraco:
```suggestion
async def test_candle_warmer_lamp_info() -> None:
```

tests/test_strip_light.py
```diff
@@ -114,13 +128,33 @@ async def test_default_info(device_case, expected_effects):
...
+
+
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
     ("basic_info", "version_info"), [(True, False), (False, True), (False, False)]
 )
-async def test_get_basic_info_returns_none(basic_info, version_info, device_case):
+async def test_get_basic_info_returns_none(
+    basic_info, version_info, device_with_candle_warmer_lamp
+):
```
bdraco:
```suggestion
) -> None:
```

bdraco (approved):

PR #457, opened 2026-03-05
feat: add roller mode for switch2pm

fankai777:
## Summary

Add roller/cover mode support for `SwitchbotRelaySwitch2PM` and fix related test cases.

### Features
- **`SwitchbotRelaySwitch2PM` now inherits from `SwitchbotBaseCover`**, enabling cover/roller control support.
- Added `open()`, `close()`, `stop()`, `set_position()` commands for roller mode.
- Added `position` and `mode` properties to `SwitchbotRelaySwitch2PM`.
- Added roller mode BLE commands: `COMMAND_OPEN`, `COMMAND_CLOSE`, `COMMAND_POSITION`, `COMMAND_STOP`.

### Parser
- `process_relay_switch_2pm`: Added `mode`, `position`, `calibration` fields to both channel 1 and channel 2 advertisement data.

### Refactor
- `SwitchbotBaseCover.__init__`: Changed `reverse` to a keyword-only argument, updated all subclass call sites (`curtain.py`, `blind_tilt.py`, `roller_shade.py`).
- Fixed `_parse_common_data` in `SwitchbotRelaySwitch`: `sequence_number` now reads from `raw_data[0]` (was `raw_data[1]`).

### Tests
- Updated `test_adv_parser.py`: Added expected `mode`, `position`, `calibration` fields to `SwitchbotRelaySwitch2PM` test cases.
- Updated `test_base_cover.py`: Fixed `SwitchbotBaseCover` instantiation to use new keyword argument signature.


switchbot/devices/relay_switch.py
```diff
@@ -213,29 +278,54 @@ async def get_basic_info(self):
         current_time_hex, current_day_start_time_hex = (
             self.get_current_time_and_start_time()
         )
-        if not (common_data := await super().get_basic_info()):
```
bdraco, line 232: There is some duplication below that could be cleaned up

bdraco:
Out of time to finish reviewing this toda. Will try to pickup reviews on this repo later this week.

switchbot/devices/relay_switch.py
```diff
@@ -204,38 +216,96 @@ class SwitchbotRelaySwitch2PM(SwitchbotRelaySwitch):
...
+    @update_after_operation
+    async def stop(self) -> bool:
+        """Send stop command to device."""
+        result = await self._send_command(COMMAND_STOP)
+        if success := self._check_command_result(result, 0, {1}):
+            self._is_opening = self._is_closing = False
+        return success
+
+    @update_after_operation
+    async def set_position(self, position: int, mode: int = 0) -> bool:
```
bdraco: This doesn't appear to honor reverse

switchbot/devices/relay_switch.py
```diff
@@ -204,38 +216,96 @@ class SwitchbotRelaySwitch2PM(SwitchbotRelaySwitch):
     def channel(self) -> int:
         return self._channel
 
+    def get_position(self) -> Any:
```
bdraco, line 234: This doesn't appear to honor reverse either

switchbot/devices/relay_switch.py
bdraco: mode is unused?

switchbot/adv_parsers/relay_switch.py
```diff
@@ -54,12 +54,18 @@ def process_relay_switch_2pm(
...
+            "position": mfr_data[14],
+            "calibration": bool(mfr_data[8] & 0b01000000),
         },
         2: {
             "switchMode": True,  # for compatibility, useless
             "sequence_number": mfr_data[6],
             "isOn": bool(mfr_data[7] & 0b01000000),
             "power": parse_power_data(mfr_data, 12),
+            "mode": (mfr_data[9] & 0b11110000) >> 4,
+            "position": mfr_data[14],
```
bdraco, line 69: Uses the same position as 1: is that correct?

bdraco (changes requested):
Please see review comments above

Onero-testdev:
Thanks for the detailed review! All issues have been addressed in the latest commit ([c62e071](https://github.com/sblibs/pySwitchbot/pull/457/commits/c62e071)):

**Copilot comments:**
- **Operator precedence bug** — Fixed: `(mfr_data[9] & 0b11110000) >> 4` now has correct parentheses.
- **Abstract methods not implemented** — Fixed: `get_position()` and `get_extended_info_summary()` are now explicitly implemented in `SwitchbotRelaySwitch2PM`.

**@bdraco comments:**
- **Code duplication** — Cleaned up: `open()`, `close()`, `stop()`, and `set_position()` have been refactored to remove duplication.
- **`set_position` doesn't honor `reverse`** — Fixed: `set_position()` now applies `position = 100 - position` when `self._reverse` is `True`.
- **`open`/`close` don't honor `reverse`** — Fixed: `open()` sends device position `100` when reversed, `close()` sends `0`, matching the base cover convention.
- **`mode` parameter unused** — Fixed: the `mode` parameter has been removed. The protocol's `0x0D 0x04` command only takes position + control-source bytes; there is no mode field.
- **Same `position` byte for channel 1 and 2** — Confirmed correct per the BLE protocol spec. In roller/cover mode, both relay channels drive a single motor together — there is only one position byte in the advertisement (Byte 21, `mfr_data[14]`). Both channels share the same physical position, so reading from the same byte is intentional.

Also added a regression test (`test_2pm_adv_short_mfr_data_returns_empty`) to guard against truncated advertisement payloads. All 1101 tests pass.

switchbot/adv_parsers/relay_switch.py
bluetoothbot: Correct for roller mode. Single motor, both channels share one position byte (`mfr_data[14]`). One caveat: advertisement uses raw `mfr_data[14]`; `get_basic_info` uses `raw_data[9] >> 1`. Confirm both same 0-100 scale.

switchbot/devices/relay_switch.py
bluetoothbot: Still partly present. `open`/`close`/`set_position` each format `COMMAND_POSITION`, send, check result. Extract `_send_position(position)` helper.

switchbot/devices/relay_switch.py
bluetoothbot: Now honored. `set_position` applies `100 - position` when reversed.

switchbot/devices/relay_switch.py
bluetoothbot: Now honored. `open` sends device 100 reversed, `close` sends 0.

switchbot/devices/relay_switch.py
bluetoothbot: Resolved. `mode` parameter removed from `set_position`.

bluetoothbot (changes requested):
Blocking issues found.

- COMMAND_OPEN / COMMAND_CLOSE are dead code
- Two different encodings for "position" (basic-info vs advertisement)

Onero-testdev:
All review feedback addressed — summary of changes in this push:

1. **Removed dead code** (`COMMAND_OPEN` / `COMMAND_CLOSE`): `open()` and `close()` both send `COMMAND_POSITION`, so the two named constants were unused. Deleted them and updated the corresponding test assertions to match the actual command sent.

2. **Fixed return type annotation** on `process_relay_switch_2pm`: was `dict[int, dict[str, Any]]`; corrected to `dict[int | str, dict[str, Any] | int]` to reflect the top-level `"sequence_number"` int value alongside the per-channel dicts.

3. **Extracted `_send_position()` helper**: `open()`, `close()`, and `set_position()` all shared the same `_send_command(COMMAND_POSITION.format(...))` + `_check_command_result(...)` pattern. Extracted into a private `async def _send_position(self, position: int) -> bool` method; motion-direction logic preserved in the callers.

4. **Merged upstream/master**: clean merge, no conflicts (35 files updated).

Tests: 55 relay-switch tests pass, 1238 total pass. ruff check + format clean.

Onero-testdev:
All review feedback has been addressed:
- Removed dead code (`COMMAND_OPEN` / `COMMAND_CLOSE`)
- Fixed the return type annotation in `adv_parsers/relay_switch.py`
- Extracted `_send_position()` helper to eliminate duplication in open/close/set_position
- Resolved merge conflicts with upstream/main — all CI checks now passing

Could you please take another look? Thanks!

bluetoothbot (changes requested):
Blocking issues found.

- Two different position encodings reach Home Assistant for the same logical value

Onero-testdev:
Fixed the position encoding inconsistency flagged by bluetoothbot.

**Root cause:** `_parse_common_data` (used by `get_basic_info`) was applying `>> 1` to the position byte, producing 0–50 instead of the correct 0–100 range. The advertisement parser (`process_relay_switch_2pm`) correctly uses the raw `mfr_data[14]` value. These two different encodings reaching Home Assistant caused the "Two different position encodings" issue.

**Fix:** Removed the `>> 1` shift — `raw_data[9]` is now used directly, matching the advertisement's encoding. Per the protocol documentation, the position is always a direct 0–100 percentage value (broadcast byte 21, `0x0F 0x71 0x0d` response, and `0x0F 0x71 0x81` response).

**Changes:**
- `switchbot/devices/relay_switch.py`: `"position": raw_data[9] >> 1` → `"position": raw_data[9]`
- Added regression test: `raw_data[9] = 0x64` (100) → asserts `position == 100`

All CI checks passing.

bluetoothbot (changes requested):
Blocking issues found.

- Position from advertisement (mfr_data[14]) and basic_info (raw_data[9]) are unverified to share the same scale
- 2PM position orientation is inverted relative to sibling covers / HA convention

PR #468, opened 2026-04-13
Fix Keypad Vision doorbell detection using seq counter instead of transient bit

zerzhang:
To fix: https://github.com/home-assistant/core/issues/167098
For the doorbell seq counter， the default value above is 0, When triggered, the loop counts from 1 to 7.


.gitattributes
```diff
@@ -0,0 +1,11 @@
+# Enforce LF line endings for all text files
```
bdraco: looks like a stray file

zerzhang: This file is intentionally added to handle cross-platform line ending issues between Windows and Unix environments.

bdraco: Please move it to a separate PR. thanks

zerzhang: fixed

bdraco (approved):

PR #472, opened 2026-04-15
Add a half lock to Lock Ultra

zerzhang:
- Add half lock


bdraco:
Traveling until Thurs 23rd so I will review when I get back or if I find time between flights

zerzhang:
> Traveling until Thurs 23rd so I will review when I get back or if I find time between flights

Thanks for the update! Safe travels — I’ll follow up when you’re back.

bdraco (approved):

PR #512, opened 2026-06-06
fix: leak detector with passive scanning

avian2:
Remove unnecessary check for data (containing service data). Service data is absent in passive scans. Later parsing code only uses mfr_data anyway so this check does not appear necessary.

I am not sure why the ckeck was there. The presence of a test suggests that it was intentional as if the sensor did not send valid mfr_data in passive scans and mfr_data should only be parsed in an active scan.

However my Switchbot leak detectors (bought July 2025 and April 2026) do send usable battery and leak bits in mfr_data also in passive scans so this check seems unnecessary and prevents me from using parse_advertisement_data in my setup.


bluetoothbot (approved):
No blocking issues found.

bdraco (approved):

bdraco:
Thanks @avian2

PR #515, opened 2026-06-10
feat(lock): add Quick Key support for Lock Ultra

mayerwin:
## Summary

Adds support for the **Quick Key** setting on the SwitchBot **Lock Ultra**. All three of its sub-settings live in a single config byte on the lock, read/written over BLE:

- `SwitchbotLock.get_quick_key()` — reads + parses the byte into `{"enabled": bool, "double_press": bool, "function": QuickKeyFunction}`.
- `SwitchbotLock.set_quick_key(*, enabled=None, double_press=None, function=None)` — a **masked write** that changes only the fields you pass (the rest keep their current value) and confirms the lock echoed the requested bits.
- A `QuickKeyFunction` enum (`LOCK_AND_UNLOCK` / `UNLOCK_ONLY` / `LOCK_ONLY`).

Scoped to `LOCK_ULTRA`; other lock models raise `SwitchbotOperationError` (the command set is untested there).

## Protocol

```
read : 57 0f 4f 04 01                         -> <status> <CFG> ...
write: 57 0f 4e 04 01 00 <mask> <value> ff     (masked write of the low nibble)
```

CFG low nibble: bit 3 = enabled, bit 2 = double-press, bits 1-0 = function.

## Testing

- **Hardware:** verified on a SwitchBot **Lock Ultra (firmware V2.6)** — every enabled / single-vs-double / function combination writes and reads back correctly.
- **Unit tests:** added to `tests/test_lock.py` (get, masked single + multi-field set, no-op guard, unsupported-model guard). The full `tests/test_lock.py` passes (203 tests).

## Note

As an interim solution I've published a Home Assistant custom integration that uses this protocol (via the existing connection) for anyone who needs it before this merges: https://github.com/mayerwin/ha-switchbot-lock-quick-key-support — it can be retired once this lands.


bluetoothbot (changes requested):
Blocking issues found.

- Undefined 4th function value crashes get_quick_key with ValueError

mayerwin:
Thanks for the review. Both points are addressed in 08a5feb:

1. **Undefined function value.** `_parse_quick_key` now checks the 2-bit function field against the set of defined `QuickKeyFunction` values before building the enum. An undefined value (e.g. `0b11` from a firmware variant or a corrupted BLE-proxy byte) logs and returns `None`, so `get_quick_key()` keeps the return-None-on-malformed contract the other parse paths follow instead of raising. I used a membership check rather than a `try`/`except` so the failure path does not emit a traceback on every poll (which would spam the HA coordinator). Added a test for that path.

2. **Export.** `QuickKeyFunction` is now re-exported from `switchbot/__init__.py` and the `const` package alongside `LockStatus`, so callers of `set_quick_key(function=...)` can import it from the package root.

Full test suite passes (1222) and ruff is clean.

bluetoothbot (dismissed):
> [!TIP]

mayerwin:
Updated in b513df5, with a field report behind it.

**Now verified on a second Lock Ultra.** A user of the interim custom component that ships this protocol ahead of the library hit a decode failure and root-caused it themselves: their lock reports the config byte as `0x8a` / `0x8e` where mine reports `0xc2` / `0xca`. Same low nibble, same behaviour, bit 6 clear. Toggling double press changed exactly the bit this PR says it should, on their hardware. So the decoding here is confirmed on two independent units rather than one, which is the thing it was missing.

That also corrected a claim in the diff. The comment said bits 7-6 are constant status flags. They are not. This PR never validated them, so the code was always correct, but the comment is now accurate and a test pins that `0x8a` and `0xca` parse identically.

Three changes came out of it:

1. **`get_quick_key()` returns the raw config byte.** That report cost several round trips purely because nothing in the stack could show the byte the lock actually sent. It was solved only when the reporter patched a log line into their own copy. Callers can surface it now, which makes the next such report a one-round one.
2. **An undefined value in the 2-bit function field no longer discards the whole reading.** The enable and trigger bits are unambiguous, so they are kept and only `function` is `None`. This follows the existing `_parse_basic_data` convention, which returns `{}` for Lock Ultra Max rather than `None`.
3. **The open review suggestion is taken.** The combined guard at the end of `set_quick_key` is split, so a truncated echo and a lock that refused a requested bit each log what happened, with tests covering both branches.

216 tests pass, ruff clean.

This has been open since June with CI green throughout, so a look from a maintainer would be welcome. Happy to rebase or split it if that helps.

One thing worth an explicit decision: `LOCK_ULTRA_MAX` is deliberately excluded from both command dicts as untested. Ultra Max support landed in #558 after this was opened, so if anyone has one and can verify, widening it is a one-line addition to each dict.

bluetoothbot (approved):
> [!TIP]

PR #488, opened 2026-05-06
Add SwitchbotKeypad device class for classic Keypad

italo-lombardi:
Adds support for SwitchBot Keypad (WoKeypad) — exposes battery level and attempt_state from BLE advertisements.
https://eu.switch-bot.com/products/switchbot-keypad

**Tested on a real device** — scanner confirmed `battery: 100` and `attempt_state` parsed correctly from live BLE advertisement data.

## Summary

The classic SwitchBot Keypad (WoKeypad) is a passive BLE-only device used to unlock doors paired with SwitchBot Lock. The advertisement parser (`adv_parsers/keypad.py`) already parses `battery` and `attempt_state` from BLE advertisements, and `SwitchbotModel.KEYPAD` is already defined, but no device class existed.

## Changes

- Add `SwitchbotKeypad` device class in `devices/keypad.py`, extending `SwitchbotDevice` (passive-only, no BLE commands)
- Export `SwitchbotKeypad` from `switchbot/__init__.py`
- Add `KEYPAD_INFO` test fixture in `tests/__init__.py`
- Add `tests/test_keypad.py` verifying battery and attempt_state are correctly parsed from advertisement data

## Test results

All 3 unit tests pass. Verified against real WoKeypad hardware — advertisement data parsed correctly.

PR #518, opened 2026-06-13
feat: implement passcode management and cloud synchronization for SwitchBot Keypad

deece:
Implement passcode management and cloud synchronization for SwitchBot Keypad

This Pull Request adds complete local and cloud passcode management and clock
synchronization capabilities for the non-vision SwitchBot Keypad (WoKeypad).

It also integrates the passive `attempt_state` advertisement property and tests proposed in PR #488 (by @italo-lombardi).

Why Cloud Sync is Needed:
- Keypad passcodes are stored and validated locally (allowing the Keypad to directly command the Lock offline without requiring internet connectivity).
- However, for security reasons, the Keypad's BLE protocol is write-only for passcodes. There is no command to read back or list saved codes over BLE.
- Since the SwitchBot app cannot query the Keypad to build its UI passcode list, it relies entirely on the SwitchBot Cloud database to track which codes exist.
- If a passcode is added locally over BLE but not synced to the cloud, it will successfully operate the Lock offline, but will be completely invisible in the official mobile app. Integrating the cloud sync option ensures the app UI stays in sync with the keypad's physical storage.

What Was Implemented:
1. Passcode Management:
   - `add_password`: Sends passcode bytes over BLE to register a new PIN code on the keypad. Parses response to retrieve the device-assigned index.
   - `modify_password`: Modifies an existing passcode by index. Overwrites values and configures active duration ranges.
   - `delete_password`: Instantly deletes a passcode from device memory by index.
   - `get_password_count`: Queries counts of stored PINs, NFC tags, fingerprints, and duress credentials.
2. Clock Synchronization (RTC):
   - `sync_time`: Synchronizes the internal device clock with millisecond precision (sending 8-byte big-endian milliseconds timestamp). Automatically scales second-level timestamps for backwards compatibility.
   - Note: Clock querying (`get_time`) was tested but found to return error 05 (unsupported) by the keypad hardware, indicating it is a write-only clock, so it has been dropped from the implementation.
3. SwitchBot Cloud Sync Option:
   - Added keyword-only parameters to `add_password`: `session`, `token`, `region`, `name`, and `creator`.
   - If provided, automatically triggers an HTTP POST request to API function 4245 to register the passcode in the SwitchBot Cloud database so it appears correctly in the official smartphone app.
4. Passive Advertisement Property Integration (from #488):
   - Exposes `attempt_state` property via `_get_adv_value` in `SwitchbotKeypad`.
   - Adds the missing `KEYPAD_INFO` test fixture to `tests/__init__.py`.

Testing Methodology:
1. Automated Tests (tests/test_keypad.py):
   - Created full coverage suite for add/modify/delete password, counts, and sync_time.
   - Implemented `test_add_password_with_cloud_sync` mocking `api_request` to verify payload structures, regional routing, and authorization headers.
   - Added advertisement parsing unit tests for battery percentage and attempt state.
   - Verified that all 14 unit tests pass cleanly.
2. Manual/Hardware Verification:
   - Tested using test_hardware.py on a physical WoKeypad and Lock.
   - Verified that syncing clock with millisecond timestamps sets the correct RTC, allowing time-limited (temporary) passcodes to evaluate successfully offline and unlock the lock.

Co-authored-by: Alastair D'Silva <alastair@d-silva.org>
Co-authored-by: Antigravity <antigravity@google.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,8 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: "3.13"
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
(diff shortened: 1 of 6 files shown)


bluetoothbot (changes requested):
Blocking issues found.

- MagicMock used in annotations but never imported — test module fails to import
- Cloud sync silently skipped on partial credentials; BLE write already committed if cloud call then fails

bluetoothbot (changes requested):
Blocking issues found.

- Orphan passcode when time-window write fails after add succeeds

bluetoothbot (dismissed):
No blocking issues found.

bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - modify_password deletes the caller's existing passcode on any failure — including failures where nothing was written
> - add_password rollback only covers protocol rejection, not the BLE exception path — orphaned always-active passcode

bluetoothbot (dismissed):
> [!TIP]

PR #521, opened 2026-06-17
fix: Infer TRV idle state when target temperature is reached with 0.5°C hysteresis

davidrule1969:
Currently, the SwitchBot TRV unconditionally reports `ClimateAction.HEATING` as long as the device is turned on, even if the room temperature has met or exceeded the target temperature. This causes smart home dashboards (like Home Assistant) to permanently show the device as actively heating.

This PR adds a 0.5°C threshold check to infer an `IDLE` state when the room is warm, matching standard thermostat behavior.

### DIFF
--- a/switchbot/devices/smart_thermostat_radiator.py
+++ b/switchbot/devices/smart_thermostat_radiator.py
@@ -190,19 +190,19 @@ class SwitchbotSmartThermostatRadiator(
         return self._get_adv_value("target_temperature")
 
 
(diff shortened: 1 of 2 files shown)


bluetoothbot (dismissed):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

bdraco (approved):

PR #543, opened 2026-07-26
feat(bot): add inverse direction controls

### LINKED ISSUE #209: [bot] inversion ON/OFF
Hi,
Do you think it is possible to support the "invert the on/off direction" when switchbot bot is in "switch mode" ?
Like the app in the picture
app

Pomme :apple:

bingxyz:
## Summary

- expose the Bot's cached inverse-direction setting
- add a setter that preserves the current switch mode and press strength
- keep the existing `is_on()` and `inverse_mode` behavior unchanged

## Why

SwitchBot Bot stores its reverse on/off direction setting on the device, but local BLE clients had no focused API to read and change it while preserving the other mode settings.

This was reproduced on a physical Bot connected directly to Home Assistant over Bluetooth: the device had its reverse-direction bit enabled, so on/off actions operated in the opposite physical direction. Clearing the setting through this new BLE API restored the expected behavior without the SwitchBot app, cloud access, or a device reset.

Following review, this PR is intentionally limited to reading and changing the persistent device setting. It no longer changes logical on/off state calculation or advertisement override behavior.

The Home Assistant entity that exposes this setting will be submitted separately after a PySwitchbot release containing this API is available.

## Testing

- upstream CI passed on Python 3.11–3.14
- local full test suite on Python 3.14: 1,306 passed
- all pre-commit hooks passed
- re-tested the scoped implementation on a physical SwitchBot Bot through Home Assistant Container

## Contribution responsibility

This implementation was AI-assisted. Responsible human: @bingxyz, who reviewed the implementation and behavior, tested it on their physical device, and authorized this submission.

Fixes #209


bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - Device-reported bit silently overrides the caller's `inverse_mode` option — breaking change for existing users
> - Inversion is applied to press-mode bots, where `isOn` is a hardcoded sentinel — yields a permanent wrong `True`
> - The core premise — that the advertised `isOn` needs flipping by the device bit — is not verified by the stated hardware testing

bingxyz:
Addressed the review in `9232db8`:

- removed device-reported inverse direction from `is_on()` state calculation
- removed the Bot-specific `update_from_advertisement()` override and related optimistic-state handling
- preserved the existing `inverse_mode` behavior
- scoped the PR to `inverse_direction()` and `set_inverse_direction()`
- re-ran the full suite and pre-commit checks, and re-tested through Home Assistant on the physical Bot

The updated CI is green. @bluetoothbot review

bingxyz:
Hi @bdraco, I addressed the bot review by removing the `is_on()` behavior changes and limiting the PR to the read/set API as suggested. CI is green and the updated implementation was tested on the physical Bot. Could you please retrigger @bluetoothbot review when convenient? Thanks!

PR #547, opened 2026-07-31
fix: standing fan night light not turning off (send 0x00 for OFF)

### LINKED ISSUE #541: SwitchbotStandingFan: NightLightState.OFF (byte 3) ignored by firmware 1.3 - should be 0
On the SwitchBot Standing Circulator Fan (device firmware **1.3**), selecting the night-light **Off** state has no effect - the RGB light stays on. The two on-levels work correctly.

## Where

`switchbot/const/fan.py` (`NightLightState`) and `switchbot/devices/fan.py` (`SwitchbotStandingFan.set_night_light()`).

## Detail

`NightLightState` encodes `LEVEL_1 = 1`, `LEVEL_2 = 2`, `OFF = 3`, and `set_night_light()` sends `57 0f 41 05 02 <value> FF FF`. On firmware 1.3 the two on levels (bytes `0x01` / `0x02`) work, but byte `0x03` is ignored - the light stays on.

The advertisement `nightLight` field is a 2-bit value that reports `0` when the light is off, so the consistent encoding is:

```
0x00 = off, 0x01 = bright (LEVEL_1), 0x02 = soft (LEVEL_2)
```

i.e. `NightLightState.OFF` should be `0`, not `3`. With `OFF = 0` the command encoding and the advertisement encoding agree, so the reported state is also correct (currently the advertised `0` can never be mapped back to a state).

## Suggested fix

- Set `NightLightState.OFF = 0` in `switchbot/const/fan.py`.
- Optionally, read the night-light level from the connection response in `get_basic_info()` so the level is available without active BLE scanning (today it is only present in the advertisement).

## Environment

- Device: SwitchBot Standing Circulator Fan with RGB Lights, firmware 1.3
- Tested on real hardware.

Originally surfaced in McMoobud/ha-switchbot-standing-fan#2.

McMoobud:
## Problem

Calling `SwitchbotStandingFan.set_night_light(NightLightState.OFF)` does not turn the night light off. The command is sent and acknowledged, but the light stays on.

The `NightLightState` enum encodes `OFF = 3`, so the OFF command goes out as byte `0x03`. On the Standing Fan (FAN2, firmware 1.3) the device silently ignores `0x03` for the night-light characteristic and only turns the light off when it receives `0x00`. `LEVEL_1` (`0x01`) and `LEVEL_2` (`0x02`) work correctly.

Fixes #541

## Fix

Scoped entirely to `SwitchbotStandingFan.set_night_light`: map `NightLightState.OFF` to `0x00` on the wire while leaving `LEVEL_1`/`LEVEL_2` untouched. The shared `NightLightState` enum is deliberately not changed, so nothing else that references it is affected.

```python
state = NightLightState(state)
value = 0 if state is NightLightState.OFF else state.value
```

## Testing

Hardware-confirmed on a physical SwitchBot Standing Fan (FAN2), firmware 1.3:
- OFF now turns the night light off reliably (previously ignored).
- LEVEL_1 / LEVEL_2 unchanged and still work.

This workaround has been running in production via a custom Home Assistant `custom_components/switchbot` fork; this PR upstreams the minimal fix.

## Notes

An alternative would be to change the enum value `OFF = 0`, but I could not verify the corresponding advertisement/parse byte for that path, so I kept the change local to the send method to avoid regressing state reporting.


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco (approved):

PR #549, opened 2026-08-03
fix(ceiling-light): decode basic info color mode bit

xtymmms1021:
## Bug Description

When a Ceiling Light reports bit 6 set in its basic-info mode byte, `get_basic_info()` returns the raw masked value `64`. Passing that value to `CeilingLightColorMode` raises:

```text
ValueError: 64 is not a valid CeilingLightColorMode
```

This can surface in consumers such as Home Assistant while updating a `WoCeiling` light.

## Root Cause

The mode field was masked with `0b01000000` but was not shifted into the enum's expected `0`/`1` range.

## Fix

- Shift the masked mode flag right by six bits.
- Add a regression case covering a basic-info byte of `0xC0`, which should decode to night color mode `1` rather than `64`.

## How to Verify

1. Run `pytest tests/test_ceiling_light.py`.
2. Confirm the `night-color-mode` case passes and returns color mode `1`.

## Test Plan

- [x] Added a regression test that failed with `assert 64 == 1` before the fix
- [x] Ceiling Light tests pass (`17 passed`)
- [x] Full test suite passes (`1301 passed`)
- [x] Ruff, formatting, and all pre-commit hooks pass
- [x] Independently reviewed with no security or logic findings

## Risk Assessment

Low — the change only normalizes a single-bit field to the enum values already used by the advertisement parser and existing color-mode mapping.

## Attribution

This pull request was prepared with assistance from Hermes Agent. Human responsible for the contribution: Taiyu Xu (@xtymmms1021).


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco (approved):

PR #557, opened 2026-08-26
fix(ceiling-light): ignore invalid color temperature

matchai:
Ceiling Light Pro can return `0xFF00` in the color-temperature field after a bare turn-on. PySwitchbot currently interprets this as `65280 K`, which consumers such as Home Assistant expose as the current color temperature.

Ignore values outside the device’s supported 2700–6500 K range. Preserve the last valid temperature when available, or use the existing 4001 K default during initial discovery.

Tests cover both cached and initial-state cases using a captured device response.


bluetoothbot (approved):
> [!TIP]

bdraco (approved):

PR #430, opened 2025-12-23
feat: add settings for Meter Pro Co2

ahrmn:
I reverse-engineered the settings of the Meter Pro (CO2) and implemented them here.
I have no means of testing if this work for the other Meter models.
An example script to fully configure a device using this would be for example
```
import asyncio

from switchbot.devices.meter import SwitchbotMeter
from switchbot.discovery import GetSwitchbotDevices
from switchbot.const import SwitchbotModel

BLE_MAC='XX:XX:XX:XX:XX:XX'
METER_MODEL=SwitchbotModel.METER_PRO_C

async def main():
    switchbotDevices = await GetSwitchbotDevices().discover()
    meter_advertisement = switchbotDevices[BLE_MAC]
    print(meter_advertisement.data)
    meter = SwitchbotMeter(device=meter_advertisement.device, model=METER_MODEL)
    await meter.show_battery_level(False)
    await meter.set_co2_thresholds(900, 1200)
    await meter.set_comfortlevel(19, 22, 40, 70)
    await meter.set_temperature_update_interval(10)
    await meter.set_co2_update_interval(10)
    await meter.set_button_function(False, False)
    await meter.set_alert_sound(False, 2)
    await meter.set_alert_co2(True, 400, 1200, False)
    
asyncio.run(main())
```

Additionally one can trigger a new measurement regardless of the frequency using 
`await meter.force_new_CO2_measurement()` and calibrate the sensor (at fresh air!) using 
`await meter.calibrate_co2_sensor()`.

PR #559, opened 2026-08-30
feat: add CO2 Settings to SwitchbotMeterProCO2

ahrmn:
Since #430 has a lot of ongoing discussions i moved a few simple parts out. In hope that this is more manageable.

This only contains some CO2-Settings of the SwitchbotMeterProCO2.
To be precise Calibrating the Sensor, forcing a new Co2 Measurement, Setting the co2 update interval and defining the thresholds for air quality.


bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - New CO2 commands ignore the device response — failures look like success

PR #563, opened 2026-09-03
feat: add OAuth support for cloud APIs

zerzhang:
Human responsible: @zerzhang.

## Summary

- Add helpers for building SwitchBot OAuth authorization URLs and exchanging authorization codes.
- Support access-token-based cloud device discovery and encryption-key retrieval.
- Validate provider responses while keeping authorization codes, access tokens, refresh tokens, and encryption keys out of logs.
- Preserve authentication and API error types through device and encryption-key requests.
- Document the public-client OAuth contract and caller responsibilities.

## OAuth wire contract

- The caller supplies a SwitchBot-issued client ID and its registered redirect URI; pySwitchbot does not embed Home Assistant credentials.
- This is a public-client flow. Open-source consumers cannot keep a client secret, so the token exchange intentionally does not send one.
- SwitchBot's current authorization server does not support PKCE. Callers must still generate, store, and validate a single-use `state`; `state` does not replace PKCE.
- The returned access token is sent directly in the `authorization` header expected by the SwitchBot internal account endpoints.
- The current Home Assistant setup flow consumes the token transiently and does not persist it, so this change does not add an unverified refresh request helper.

## Error semantics

HTTP 401 and 403 responses now surface as `SwitchbotAuthenticationError`, including on the pre-existing password flow. Other provider API failures surface as `SwitchbotApiError`; transport and service-availability failures use `SwitchbotAccountConnectionError`.

## Testing

- `poetry run pytest --cov=switchbot tests`
- 1403 passed
- `switchbot/oauth.py` and the new shared request-ID helper are fully covered


bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Token exchange sends no client authentication and no PKCE — and the helper signature can't express either
> - Unknown-model diagnostic loses the payload that exists to add new device support

zerzhang:
Thanks for the detailed review. I addressed the diagnostic and error-handling findings in `2d7dbce` and clarified the OAuth contract in the README and PR description.

On the OAuth wire contract: there is not currently a public SwitchBot OAuth document I can link. The implementation follows the SwitchBot-provisioned public-client flow used by the current consumer. `client_id` and the registered `redirect_uri` are supplied by the caller and are not embedded in pySwitchbot. Because the consumers are open source, a `client_secret` cannot be kept confidential and is intentionally not sent. SwitchBot's current authorization server also does not support PKCE, so I documented that limitation rather than exposing parameters the server cannot use. The caller must generate and validate a single-use `state`; the README explicitly notes that state does not replace PKCE.

Changes made from the review:

- Restored the original full unknown-model payload and unmasked MAC diagnostic.
- Preserved `SwitchbotAuthenticationError` through the later device-list and encryption-key requests.
- Preserved `SwitchbotApiError` from the key endpoint and classify malformed `communicationKey` responses as API errors instead of connection errors.
- Documented the 401/403 exception-type change for the existing password flow.
- Parse only bounded `error` and `error_description` fields from OAuth failures, redact an echoed authorization code, and never log the raw response.
- Moved the duplicated request-ID extraction into a shared helper.
- Relaxed exact debug-message assertions while retaining the negative secret-leakage assertions.
- Added tests for the new authentication, API, connection, malformed-response, and OAuth error paths.

The token returned by this flow is expected directly in the SwitchBot internal account endpoints' `authorization` header, without a `Bearer` prefix. The current Home Assistant setup flow uses the token transiently and does not persist it. I therefore did not add a refresh helper without a verified refresh wire contract.

Local result: `poetry run pytest --cov=switchbot tests` — 1403 passed; the OAuth module and shared request-ID helper are fully covered.

bluetoothbot (dismissed):
> [!TIP]

bdraco:
@zerzhang You should have access to merge this once the bot approves.  There are some nits above to address though

bluetoothbot (approved):
> [!TIP]

PR #568, opened 2026-09-16
feat(art_frame): add random_image method

italo-lombardi:
## Summary

Adds a `random_image()` method to `SwitchbotArtFrame`, mirroring the existing `next_image()` / `prev_image()` helpers. It selects a random image other than the currently displayed one and sends the set-image command over BLE.

## Why here

This started as inline logic in a Home Assistant PR that adds a "random image" button for the Art Frame: home-assistant/core#182123. A maintainer pointed out that the random-selection logic is an integration-level abstraction and should live in the library instead, so downstream consumers call one method rather than reimplementing the selection. This PR moves it here; the HA PR will be updated to call `random_image()` once this is released.

## Behavior

- Picks a random image ID from `all_images_index`, excluding the current image.
- Raises `RuntimeError` when fewer than two images are available (consistent with `_select_image_index`).

## Testing

- `test_random_image` — parametrized, asserts the sent command is one of the non-current image IDs.
- `test_random_image_with_single_image` — asserts `RuntimeError` with a single image.

All existing art-frame tests continue to pass.

> [!WARNING]
> **Not yet verified on real hardware.** Only the mocked unit tests above have run. I don't currently have a way to test `random_image()` against a physical Art Frame over BLE. Help testing on a real device (confirming the displayed image actually changes and never repeats the current one) would be appreciated before merge.


italo-lombardi:
@bluetoothbot please review

italo-lombardi:
@bluetoothbot review

